PR #3, opened 2019-03-14
Changed rbac

varyumin:
I think this will not work in Kubernetes. `apiGroups: ["quota.openshift.io"]` is API group only in OpenShift


cakrit:
Just checking, you are able to merge it, right? If not, I'll need to look at the settings...

varyumin:
Yes, I can. Thx

PR #11, opened 2019-04-02
* Fix RBAC. ClusterRoleBinding installed SeviceAccount only from name…

varyumin:
* Fix RBAC. `ClusterRoleBinding` can use `SeviceAccount` only from namespace `default`. If you installed in another namespace ("not `default`") then `ServiceAccount` and `ClusterRoleBinding` didn't work
* Feature RBAC. You can pick between "turn on" (`helm ... -set rbac.create=true`) or "turn off" (`helm ... -set rbac.create=false`) RBAC.
* Feature RBAC. You can create serviceAccount or use your existing `serviceAccount`.
* I have changed liveness and readiness probes on `/api/v1/info`
* Added myself to Chart.yaml (if you don't mind @cakrit )

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,5 @@
+# IntelliJ project files
+.idea
+*.iml
+out
+gen
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -6,3 +6,5 @@ description: Real-time performance monitoring, done right! https://my-netdata.io
 maintainer:
   name: Chris Akritidis
   email: chris@netdata.cloud
+  name: Vladimir Ryumin
+  email: vryumin@gmail.com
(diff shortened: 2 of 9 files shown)


varyumin:
Ok. I will update README.md

cakrit:
I also mentioned you in https://github.com/netdata/netdata/issues/5764. 
The probes just look for an HTTP 200 code, right? Just so I know if we have any dependency on getting that right. e.g. can they look for a specific answer in that response? It's so I know how to design it, because I'll put in some more logic on when and how netdata responds.

cakrit:
Ok, I upgraded and saw the slave containers getting restarted serially. It takes ~1 minute for the last of my 3 slaves to get upgraded, but I suppose this is normal. I'll approve as soon as you update the README. 

I noticed another problem while testing, I'll create a new issue for that.

varyumin:
> The probes just look for an HTTP 200 code, right? 
If we used `httpGet` probes, Yes it will be `200 OK`

varyumin:
I have updated README.md

cakrit:
Thanks. Merge whenever you want.

PR #13, opened 2019-04-05
Use persistent GUIDs per hostname

### LINKED ISSUE #12: Slave pods should use persistent identifiers per node
I just noticed in `/var/cache/netdata` that we lose the history of each slave node when we do a `helm upgrade`. This certainly has to do with the way the slaves identify themselves to the master and we need to fix it.

cakrit:
Fixes #12 

Made the master machine GUID persistent as well, in case anyone wants to stream master data to another netdata somewhere.

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,6 +1,6 @@
 name: netdata
 home: https://github.com/netdata/netdata
-version: 0.0.7
+version: 0.0.8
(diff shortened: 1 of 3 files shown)


cakrit:
@varyumin I tested this, but I'd appreciate your review as well.

varyumin:
I have tested! Well done!

PR #15, opened 2019-04-08
Add serviceName to statefulset spec

### LINKED ISSUE #14: Stateful set requires serviceName to work
Trying to deploy I am getting the following error:

```bash
$ kubectl apply -f statefulset.yaml
```

> error: error validating "statefulset.yaml": error validating data: ValidationError(StatefulSet.spec): missing required field "serviceName" in io.k8s.api.apps.v1beta2.StatefulSetSpec; if you choose to ignore these errors, turn validation off with --validate=false

According to the [spec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/#statefulsetspec-v1-apps) this field is missing from the helm chart template.

cakrit:
Fixes #14 

Added serviceName to statefulset.yaml spec.

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,6 +1,6 @@
 name: netdata
 home: https://github.com/netdata/netdata
-version: 0.0.8
+version: 0.0.9
 appVersion: v1.13.0
 description: Real-time performance monitoring, done right! https://my-netdata.io/
 maintainer:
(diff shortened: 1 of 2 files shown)


varyumin:
It looks perfect! `serviceName` needs to have `statefulSet` of course! I'm curious, but would we use `serviceName` for `slave`?

cakrit:
I didn't see serviceName in the spec for daemonset in the official docs, so I left it out

PR #18, opened 2019-04-29
add sysctl init container

kam1kaze:
Hi,

I need to set `vm.max_map_count` sysctl param to bypass the `Cannot allocate memory` issue. So I added the init container, where sysctl params could be managed.

Thanks

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,6 +1,6 @@
 name: netdata
 home: https://github.com/netdata/netdata
-version: 0.0.9
+version: 0.0.10
 appVersion: v1.13.0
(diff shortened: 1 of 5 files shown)


cakrit:
Are you allowed to press the merge button @kam1kaze? I still haven't learned the github rules on who can do it and I don't want to steal your glory if you can! :)

ilyam8:
> I still haven't learned the github rules on who can do it

i guess only persong with write access.

___

Not related, but, i noticed you are still using 1.13.0

kam1kaze:
@cakrit I could not manage such PR.

PR #24, opened 2019-05-16
k8s monitoring configuration

### LINKED ISSUE #21: Add parameters to values.yaml to enable k8s monitoring
netdata v1.14 has [coredns](https://docs.netdata.cloud/collectors/go.d.plugin/modules/coredns/), [kubeproxy](https://docs.netdata.cloud/collectors/go.d.plugin/modules/k8s_kubeproxy/) and [kubelet](https://docs.netdata.cloud/collectors/go.d.plugin/modules/k8s_kubelet/) collectors. 
We need to make enabling them via the helm chart much easier, i.e. add the config files for them and probably provide default values for their respective `metrics` endpoints.

cakrit:
Fixes #21 

The stock configs already enabled k8s monitoring for kubelet, kubeproxy and coredns, but added the option of overriding the settings in values.yaml. 

Other minor changes in this PR:
- Add a newline at the end of each config file (actually with some spaces on the last line too, but I couldn't find a way to avoid it).
- Fix the tag in values.yaml to v1.14.0

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,6 +1,6 @@
 name: netdata
 home: https://github.com/netdata/netdata
-version: 0.0.11
+version: 0.0.14
 appVersion: v1.14.0
 description: Real-time performance monitoring, done right! https://my-netdata.io/
 maintainer:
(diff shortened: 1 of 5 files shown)


cakrit:
You're too nice @varyumin , thanks for your kind words.

Issue #25, opened 2019-05-18
env field at wrong position in DaemonSet and StatefulSet


karuppiah7890:
Using the latest version v2.14 of helm, which has manifest validation feature, installation will lead to this error:

```
$ helm install --name test . --dry-run
Error: error validating "": error validating data: ValidationError(DaemonSet.spec.template.spec): unknown field "env" in io.k8s.api.core.v1.PodSpec
```

on fixing daemonset, it tells about statefulset error too

```
$ helm install --name test . --dry-run
Error: error validating "": error validating data: ValidationError(StatefulSet.spec.template.spec): unknown field "env" in io.k8s.api.core.v1.PodSpec
```

I will push a fix for this when I raise a PR for #17 😄 

Also, extra info - there's a bug in 2.14 which doesn't allow upgrading releases which have bad manifests (like this one, with env in wrong position) even though the latest chart has the correct manifests (which passes validations). If people are using netdata chart and are facing any issues after upgrading to helm v2.14 , then we need to ask them to use latest version of the chart (after the PR fix) and also use canary version helm and tiller, or ask them to wait for v2.14.1 and just use latest version of chart, and everything will still work, and releases will also get upgraded with proper manifests

PR #26, opened 2019-05-18
Persistence toggle and fixes

### LINKED ISSUE #17: Provide persistence option in values.yml
There should be an option to create no persistent storage at all, e.g. when using the master instance for prometheus scraping only.

karuppiah7890:
Fixes #17 and #25 

Summary of the features and fixes
1. Support toggling of persistence volumes for database and alarms
2. Fix `master.alarms.storageclass` being used for claiming volume for database too
3. Add apiVersion in Chart.yaml as it's required in helm3 and canary version of helm2 also shows linting error for the same
4. Fix `maintainers` key in Chart.yaml
5. Fix env field being in pod spec instead of inside container spec, in both DaemonSet and StatefulSet

cc @cakrit @varyumin . Please review 😄

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: netdata
 home: https://github.com/netdata/netdata
-version: 0.0.12
+version: 0.0.13
 appVersion: v1.14.0
 description: Real-time performance monitoring, done right! https://my-netdata.io/
 maintainers:
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,10 +1,13 @@
+apiVersion: v1
 name: netdata
 home: https://github.com/netdata/netdata
 version: 0.0.12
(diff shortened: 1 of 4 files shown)


karuppiah7890:
Sure! 😄  And I put two commits. Should I change the Chart version in both the commits? Then the version would be 0.0.14 . One is a feature and the other is a fix. So I think two version bumps is okay

cakrit:
No need to make it .14, it's just one PR, resulting in a new version. 
I'll be changing to 1.0.0 anyway tomorrow or on Monday, after I put in the tag 1.15.0

karuppiah7890:
Pushed the appropriate changes 😄 version is 0.0.13

cakrit:
And it's in! Thanks @karuppiah7890 !!!

karuppiah7890:
🎉 Yay! Thanks @cakrit 😁

PR #29, opened 2019-05-20
add traefik ingress option

kareem-elsayed:


kareem-elsayed:
Update now in README to be set as a value

PR #30, opened 2019-05-20
add ability to manage custom configs

kam1kaze:
Hi,

I don't like to modify the source code of the helm package, if I need to enable some plugin. So I add such possibility.

Thanks

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: netdata
 home: https://github.com/netdata/netdata
-version: 0.0.13
+version: 1.0.0
 appVersion: v1.14.0
 description: Real-time performance monitoring, done right! https://my-netdata.io/
 maintainers:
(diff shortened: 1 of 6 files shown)


kam1kaze:
I thought about this, but it is a breaking change for the helm package. But if it's ok for you, I'll update the PR.

cakrit:
Make the version 1.0.0 then. It's much better and cleaner the way you did it. We just need to be careful to change the major version for every breaking change.

cakrit:
Merging and I'll do some more updates to the README, along with the change to include v1.15.0.

kam1kaze:
@cakrit ок, thanks. I forgot about README :)

cakrit:
Don't worry about it, great job mate. I'm adding you to the maintainers too in Charts.yaml. If you like, drop me an email so I can add your email too (chris@netdata.cloud), otherwise I'll just leave your name there (hoping it doesn't break helm).

PR #39, opened 2019-07-24
Move apparmor annotation to pod template

charlieegan3:
Based on examples here: https://kubernetes.io/docs/tutorials/clusters/apparmor/ I think this annotation needs to be on the pods themselves to get through PSP checks.

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: netdata
 home: https://github.com/netdata/netdata
-version: 1.1.1
+version: 1.1.2
 appVersion: v1.16.0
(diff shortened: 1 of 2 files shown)


cakrit:
@josecv can you check this? It will be quicker than me trying to understand how these things work...
Will wait for a couple of days for a response and merge if I don't.

cakrit:
But @charlieegan3 one thing that definitely needs to be done, is for the chart version to be increased in `Chart.yaml`.

charlieegan3:
I bumped the patch version, but :+1: to someone confirming this (I needed to do this in our cluster but a second pair of :eyes: appreciated)

Issue #40, opened 2019-08-09
Apparmor annotation still not taking affect


joell:
https://github.com/netdata/helmchart/blob/c0fa186ddcbd9be488309dbffb85d9f57b521ab8/templates/daemonset.yaml#L20

This annotation did not take affect in my case.  First off, it seems that the second `annotations` clause for `checksum/config` on line 28 below seems to stomp over the apparmor annotation.

After merging the two `annotations` clauses, I then got errors from K8S because the annotation reference a container name not found in the pod.  (The pod is named `netdata-slave` but the container within that pod is named `netdata`.)  I found it was necessary to change `/netdata-slave` to `/netdata`, after which everything worked.

In summary, the following change got things working for me:
```diff
--- a/templates/daemonset.yaml
+++ b/templates/daemonset.yaml
@@ -17,7 +17,8 @@ spec:
   template:
     metadata:
       annotations:
-        container.apparmor.security.beta.kubernetes.io/netdata-slave: unconfined
+        container.apparmor.security.beta.kubernetes.io/netdata: unconfined
+        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
       labels:
         app: {{ template "netdata.name" . }}
         release: {{ .Release.Name }}
@@ -25,8 +26,6 @@ spec:
 {{- with .Values.slave.podLabels }}
 {{ toYaml . | trim | indent 8 }}
 {{- end }}
-      annotations:
-        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
 {{- with .Values.slave.podAnnotations }}
 {{ toYaml . | trim | indent 8 }}
 {{- end }}
```

cakrit:
@josecv and @charlieegan3 any comments on this? Should we do another PR to make the changes above?

josecv:
@cakrit yup, with some minor adjustments -- basically the `with .Values.slave.podAnnotations` needs to be moved to be under checksum/config so that all annotations get merged together:

```
        container.apparmor.security.beta.kubernetes.io/netdata: unconfined
        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
 {{- with .Values.slave.podAnnotations }}
 {{ toYaml . | trim | indent 8 }}
 {{- end }}
```

I'd also recommend having a way to turn off the appArmor annotation btw

cakrit:
[...]

PR #41, opened 2019-08-09
Correctly set DaemonSet AppArmor annotation

charlieegan3:
In https://github.com/netdata/helmchart/commit/c0fa186ddcbd9be488309dbffb85d9f57b521ab8 I missed that there was an existing annotations block.

This also makes the AppArmor annotation configurable (though on by default).

Fixes https://github.com/netdata/helmchart/issues/40

Signed-off-by: Charlie Egan <charlieegan3@users.noreply.github.com>

### DIFF
--- a/templates/daemonset.yaml
+++ b/templates/daemonset.yaml
@@ -17,17 +17,18 @@ spec:
   template:
     metadata:
       annotations:
+      {{- if .Values.slave.podAnnotationAppArmor.enabled }}
         container.apparmor.security.beta.kubernetes.io/netdata-slave: unconfined
+      {{- end }}
+        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+{{- with .Values.slave.podAnnotations }}
+{{ toYaml . | trim | indent 8 }}
+{{- end }}
       labels:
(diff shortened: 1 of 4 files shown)


templates/daemonset.yaml
```diff
@@ -17,18 +17,19 @@ spec:
   template:
     metadata:
       annotations:
+      {{- if .Values.slave.podAnnotationAppArmor.enabled }}
         container.apparmor.security.beta.kubernetes.io/netdata-slave: unconfined
```
joell: This should still produce an error if you enable the AppArmor annotation and try to deploy it.  The default name of the container in the pod will be `netdata`, but the container name given here is `netdata-slave`.  It's necessary to use the *container's* name here, not the pod's name.

charlieegan3: Oops, good spot - I've set it to the chart name in 5bf8a11483cca54ba0b9deda277af49c9a90e8d4

charlieegan3:
Well remembered, should be in dfffebfe97fbbd67bae7a444dfda6fd5da4cb087

cakrit:
Excellent @charlieegan3, thanks!

PR #58, opened 2019-11-03
Use by default the default storage class for PVCs

ktsakalozos:
I think using the default storage class is a more sane default value for the `storageclass`.

Also, since the empty string as storage class has a special meaning we use "-" to denote the absence of storageclass (resulting to selecting the default storageclass) [1]

[1] https://github.com/helm/helm/issues/2600


cakrit:
I like this solution. Thanks!

PR #63, opened 2019-11-09
Optionally disable daemonset

huzlak:
**Reasoning:**   
I'd like to install only master with this helmchart via helmfile together with rest of the components which are installed with helmfile. Currently I have to template chart with helm and remove slave configmap and daemonset files. I run netdata (close to our prometheus instance which is then scraping the netdata) in managed K8s cluster where I don't have possibility to access worker nodes data as those have their own managed monitoring solution so there's no reason to deploy the slave daemonset. I'd like to use it as streaming master for netdata instances running in openstack VMs.


ncmans:
Thanks @huzlak for the PR. LGTM.

PR #67, opened 2019-12-20
ingress.yaml: change the yaml to ref the new API

dougsland:
Reference: https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/

### DIFF
--- a/templates/ingress.yaml
+++ b/templates/ingress.yaml
@@ -2,7 +2,7 @@
 {{- $fullName := include "netdata.name" . -}}
 {{- $ingressPath := .Values.ingress.path -}}
 
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1beta1
 kind: Ingress
 metadata:
   name: {{ $fullName }}


templates/ingress.yaml
```diff
@@ -2,7 +2,7 @@
 {{- $fullName := include "netdata.name" . -}}
 {{- $ingressPath := .Values.ingress.path -}}
 
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1beta1
```
ncmans: The new API has only been [available since 1.14](https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/) which is still quite recent (e.g. GKE stable as of today runs on 1.13). The old API is going away in 5 future versions from now (in 1.20). We should therefore delay this upgrade closer to removal of the API to keep as much backward compatibility as possible.

ncmans:
Thanks @dougsland for your contribution. As I noted in the [line comment](https://github.com/netdata/helmchart/pull/67#pullrequestreview-335532824), it's a bit early to upgrade to the new API which has only been available since 1.14.

dougsland:
I am wondering why not use branches for each version of kubernetes and leave master as the sync with the last bits? 

Example: 
    1.16 -> branch 1.16
    1.14 -> branch 1.14
    master -> most recent development compatible.

@ncmans  ^

dougsland:
Closing, no attention here.

knatsakis:
@dougsland thank you very much for your proposal. It sounds interesting to me and I would like to discuss it with @ncmans.

Please give us some time to think about this.

Sorry for not giving an answer yet but it's been mostly vacations causing this delay.

ncmans:
> I am wondering why not use branches for each version of kubernetes and leave master as the sync with the last bits?
> 
> Example:
> 1.16 -> branch 1.16
> 1.14 -> branch 1.14
> master -> most recent development compatible.

@dougsland 
It's more maintenance effort for us and thus we need to justify it. Is there any particular benefit we will get by using the newer API version (besides getting out of deprecated land of course)?

cakrit:
Why not just make them variables, configurable from `values.yaml`?

cakrit:
> Why not just make them variables, configurable from `values.yaml`?

Stupid idea by itself. API versions may support different things. So essentially, it would need to be combined with some `if` statements in the config files. Then we avoid the need for branches and we can support multiple API versions at the same time.

dougsland:
> > I am wondering why not use branches for each version of kubernetes and leave master as the sync with the last bits?
> > Example:
> > 1.16 -> branch 1.16
> > 1.14 -> branch 1.14
> > master -> most recent development compatible.
> 
> @dougsland
> It's more maintenance effort for us and thus we need to justify it. Is there any particular benefit we
> will get by using the newer API version (besides getting out of deprecated land of course)?

You guys don't need to update your code with the recent changes in the API (or create branches). It's open source project and someone could fork and make the update. However, don't expect all users will migrate to recent version of Kubernetes and will carry easily your bits. (For me at least, without the patch, breaks and in my case, I had time to dig and fix it).

ncmans:
> For me at least, without the patch, breaks and in my case, I had time to dig and fix it

What broke? Can you elaborate please?

dougsland:
> > For me at least, without the patch, breaks and in my case, I had time to dig and fix it
> 
> What broke? Can you elaborate please?

yaml:
apiVersion: extensions/v1beta1

PR #70, opened 2020-01-26
Support for static loadbalancerIP

MansM:
I need to be able to set a static loadbalancer IP so it is reproducable. Probably others as well ;-)


MansM:
@cakrit done

cakrit:
Thanks for the contribution @MansM!!!

PR #87, opened 2020-04-14
Bump NetData Helm Chart from v1.21.0 => v1.21.1

prologic:
Also adds a tool to do this some-what automatically until we trust that the
tool does the right thing 100% of the time; then we'll integrate this into a
GHA workflow.

See `./tools/tools/update_helmchart.sh`. Usage is simple:

```#!sh
$ ./tools/tools/update_helmchart.sh
```

Which will bump the `patch` version by default.

Otherwise:

```#!sh
$ ./tools/tools/update_helmchart.sh v1.22.0
```

To update to a new version entirely.

The functions herein are 100% POSIX Sh compliant and the version parsing and
handling supports `x.y.z` as well as optional build `x.y.z-b` and bumping
any component with resetting others to zero as appropriate (_untested_).

### DIFF
--- a/tools/update_helmchart.sh
+++ b/tools/update_helmchart.sh
@@ -110,7 +110,13 @@ _main() {
     new_version="$1"
   fi
 
-  new_appVersion="$(bump_version "$old_appVersion")"
+  # Bump patch version by default
+  if [ -z "$1" ]; then
+    new_appVersion="$(bump_version "$old_appVersion")"
+  else
+    # Or use supplied new version from first argument
+    new_appVersion="$1"
+  fi
 
   printf "\n"
   printf "New Chart version:    %s\n" "$new_version"
--- a/tools/update_helmchart.sh
+++ b/tools/update_helmchart.sh
@@ -1,0 +1,133 @@
+#!/bin/sh
+
+save_argv() {
+  for i; do
+    printf %s\\n "$i" | sed "s/'/'\\\\''/g;1s/^/'/;\$s/\$/' \\\\/"
+  done
+  echo " "
+}
+
+restore_argv() {
+  eval "set -- $*"
+}
+
+fnmatch() {
+  case "$2" in
+    $1)
+      return 0
+      ;;
+    *)
+      return 1
+      ;;
+  esac
+}
+
+parse_version() {
+  r="${1}"
+  s="${2}"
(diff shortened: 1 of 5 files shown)


tools/update_helmchart.sh
```diff
@@ -0,0 +1,133 @@
+#!/bin/sh
```
knatsakis, line 1: I don't think it helps aiming for POSIX compatibility in this script, since it is meant to run in an environment under our control. I think it just hurts readability.

Also, regarding version parsing:
- Maybe we shouldn't reinvent the wheel and instead use something like this: https://github.com/cloudflare/semver_bash, then again maybe it's not a good idea to rely on external tools for this. I don't know.
- It would have been simpler if you could just pass the new version as an argument. I am not sure in which situation you would want this script to calculate the new version by itself.

prologic: > I don't think it helps aiming for POSIX compatibility in this script, since it is meant to run in an environment under our control. I think it just hurts readability.

Yeah whilst ai would normally agree with you; I don't really want to pull in an external dep here. I also based this off of existing code we wrote for the NetData Installer but I'm improving it overall so we can use it in the Static installer too -- It _has_ to be POSIX or things don't really work out for folks on smaller/embedded systems without Bash.

> Also, regarding version parsing:
> 
> * Maybe we shouldn't reinvent the wheel and instead use something like this: https://github.com/cloudflare/semver_bash, then again maybe it's not a good idea to rely on external tools for this. I don't know.

See comment above :)

> * It would have been simpler if you could just pass the new version as an argument. I am not sure in which situation you would want this script to calculate the new version by itself.

I can add this capability easily. Normally we just bump the version though and this is useful to maintain the `version` of the helm-chart itself in lock-step. I will add the capability of passing in an arbitrary value for `appVersion` though in a new commit.

prologic: > I will add the capability of passing in an arbitrary value for appVersion though in a new commit.

Done

knatsakis: > I also based this off of existing code we wrote for the NetData Installer but I'm improving it overall so we can use it in the Static installer too

If it's based on existing code, :+1: for reusability.

prologic:
@ncmans Can I get a review/approval here so I can merge this already? :)

PR #91, opened 2020-04-30
Add priority class name

ncmans:
This is useful for infrastructure components and especially monitoring that needs to be there when things go wrong. Without a custom priority class, netdata agents can be killed under resource pressure just when you need them to be recording stats.

[Some](https://github.com/helm/charts/blob/6c7638eef1730a6e4de48f37fa3262a427106885/stable/prometheus/values.yaml#L48) other [charts](https://github.com/helm/charts/blob/6c7638eef1730a6e4de48f37fa3262a427106885/stable/datadog/values.yaml#L454) with this config.

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: netdata
 home: https://github.com/netdata/netdata
-version: 1.1.14
+version: 1.2.0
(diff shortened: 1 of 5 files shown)


ilyam8:
This is a very nice option!

PR #96, opened 2020-05-05
Fix DNS errors

### LINKED ISSUE #95: DNS errors with HostNetwork and ClusterFirstWithHostNet settings
Hi All @cakrit @ilyam8  @zack-shoylev @thiagoftsm @vlvkobal 

Here is what we found to debug connectivity and streaming problem:
The Helm chart installs netdata with dnsPolicy: ClusterFirstWithHostNet, which causes DNS errors, making slave - master streaming haphazard (even without encrypting streaming, or influxdb backend). Setting it to ClusterFirst solves the problem, but we are afraid that we lose some of the metrics.

_Originally posted by @Peggy4444 in https://github.com/netdata/netdata/issues/8847#issuecomment-623980514_

cakrit:
Fixes #95 

Removed HostNetwork and ClusterFirstWithHostNetsettings, to let them go to default. 
Installed it in our test k8s cluster and didn't see any issues, but we'll need some more info on why this was done in the first place.

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: netdata
 home: https://github.com/netdata/netdata
-version: 1.2.0
+version: 1.2.1
 appVersion: v1.21.0
 description: Real-time performance monitoring, done right! https://my-netdata.io/
 maintainers:
--- a/templates/daemonset.yaml
+++ b/templates/daemonset.yaml
@@ -37,7 +37,6 @@ spec:
       restartPolicy: Always
       hostPID: true
       hostIPC: true
-      hostNetwork: true
       {{- if .Values.slave.priorityClassName }}
       priorityClassName: "{{ .Values.slave.priorityClassName }}"
       {{- end }}
@@ -183,5 +182,4 @@ spec:
         - name: nodeuid
           emptyDir: {}
         {{- end }}
-      dnsPolicy: ClusterFirstWithHostNet
 {{- end }}


cakrit:
cc @varyumin @Peggy4444

cosmix:
Hmm, I haven't tested this, but my understanding is that any collector that requires access to the host network namespace, including the loopback device, to collect data won't work if you remove `hostNetwork`. That might also be useful to gather data from other pods running on that k8s node.

cakrit:
We mount the host's `/proc`, we don't need any other info to show network interface stats. I don't know if these settings will have any effect on monitoring the applications in other pods, I don't quite understand the setting myself.

ncmans:
Is there any metrics collected from applications via their services listening on the localhost? If so, then `hostNetwork` is needed in order to connect to them.

https://kubernetes.io/docs/concepts/policy/pod-security-policy/#host-namespaces

cakrit:
> Is there any metrics collected from applications via their services listening on the localhost? If so, then `hostNetwork` is needed in order to connect to them.
> 
> https://kubernetes.io/docs/concepts/policy/pod-security-policy/#host-namespaces

The current helmchart can't monitor other services now anyway, because netdata doesn't know how to find them, until @ilyam8 's autodetection work is completed.

@ilyam8 will autodiscovery return "localhost" or IPs that are going to be accessible from our pod? 
@ncmans how do we access our Redis etc. in our cloud infra?

prologic:
I'm not sure that mounting `/proc` is sufficient to collect metrics from the Host's networking stack. Normally a container is often setup with a different network namespace entirely than that of the parent pid that spawned it. This means it should in theory have zero access to other network namespaces _unless_ it has been explicitly granted ot attached to. I could be wrong though.

cakrit:
> I'm not sure that mounting `/proc` is sufficient to collect metrics from the Host's networking stack. Normally a container is often setup with a different network namespace entirely than that of the parent pid that spawned it. This means it should in theory have zero access to other network namespaces _unless_ it has been explicitly granted ot attached to. I could be wrong though.

No, we get container interfaces as well and even link them to the containers, so they appear under the correct place on the right menu.

prologic:
> No, we get container interfaces as well and even link them to the containers, so they appear under the correct place on the right menu.

Yeah okay. Bypassing the namespace entirely and bind-mounting host volumes seems to get this information into the slave.

Removing `hostNetwork: true` seems somehow wrong ot me but if the slave doesn't need access to the host network at all because we bind-mount all the host resources we want to monitor fine 🤷‍♂️I'm not sure this really explains the DNS issues though

ncmans:
Quite incidentally, I may have come across an issue arising from use of `hostNetwork` as well. In may be a complete red-herring but let's keep this open while I confirm if it's in fact related.

Specifically, while migrating pods across node pools, the migrated ones couldn't connect to some of the services due to DNS issues and looking around, the calico networking pods were in pending state due to error: `node(s) didn't have free ports for the requested pod ports`.

[Digging around](https://www.google.com/search?q=%22didn%27t+have+free+ports+for+the+requested+pod+ports%22), it seems related to `hostNetwork: true`.

ilyam8:
> Is there any metrics collected from applications via their services listening on the localhost?

kubelet

prologic:
> > Is there any metrics collected from applications via their services listening on the localhost?
> 
> kubelet

Then you need Host networking.

cakrit:
@prologic found a very nice issue, see https://github.com/kubernetes/kubernetes/issues/87852 and specifically https://github.com/kubernetes/kubernetes/issues/87852#issuecomment-585254957
That issue references https://github.com/coreos/flannel/issues/1243

Sounds like we're doing what we should be doing here. I believe we need to close this PR and then close the issue with references to these two, plus the explanation that the workaround of disabling these settings means we won't be able to monitor kubelet, coreDNS and kube-proxy, right @ilyam8?

varyumin:
If you want, get access to external service  "ClusterFirstWithHostNet" then dns first try to resolve a hostname from resolve.conf and after will try to resolve inside kubernetes. If you don't have external services, you have to use the default policy
https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy

cakrit:
Actually, we should probably just make these settings configurable via values.yaml. 
@varyumin I don't quite get what you just wrote. What would be in `resolv.conf` so that `netdata` can reliably resolve to the master's IP?

varyumin:
If DNS policy equal ClusterFirstWithHostNet application inside a cluster try to resolve hostname from dns servers wich in resolv.conf. In another action, app tries to resolve a hostname in inside cluster kubernetes like, kube-dns or core-dns

ilyam8:
>  we won't be able to monitor kubelet, coreDNS and kube-proxy

I think the problem with kubelet only, because it is the only no containerized thing. 

> Then you need Host networking.

i assume we can access it using apiserver

`/api/v1/nodes/<NODE>/proxy/metrics`

ilyam8:
related https://github.com/netdata/netdata/issues/8060

ilyam8:
To user kubelet `/pods` endpoint for name resolution (cgroups sh script) would be handy too, much less requests to apiserver.

ilyam8:
> we won't be able to monitor kubelet, coreDNS and kube-proxy

yes, with [current implementation](https://github.com/netdata/helmchart/blob/955b02602ffe6f807d80e2f88f45f68760a2badc/values.yaml#L188-L214) - yes

cakrit:
We will make these settings configurable via `values.yaml`

PR #97, opened 2020-05-11
Bump NetData Helm Chart from v1.21.1 => v1.22.0

Ferroin:


lgg42:
Thanks for making the PR so fast!
Is it working for you? I tried it and after deployed I'm getting the following message:
`Access to file is not permitted: /usr/share/netdata/web//index.html`

command: `helm upgrade netdata netdata/netdata -i --atomic --namespace kube-system -f values/netdata.yaml`

I'm still using the chart v1.21.0 but I changed the image tag in values/netdata.yaml
```
image:
  repository: netdata/netdata
  tag: v1.22.0
  pullPolicy: Always
```

I also tried purging the release (and deleting remaining objects like PVC and PV) and installing it from zero, got the same result.

Ferroin:
@lgg42 I'm fairly certain that the issue you're seeing is actually in our Docker image and not the Helm chart. There's a new release in progress right now that fixes that issue, and there should be a new PR here for that some time within the next 24 hours.

Ferroin:
Closing this PR as we've decided to skip v1.22.0 in the Helm chart due to the usability issues. v1.22.1 should have a PR up soon.

PR #98, opened 2020-05-12
Bump NetData Helm Chart from v1.21.1 => v1.22.1

Ferroin:


Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 name: netdata
 home: https://github.com/netdata/netdata
-version: 1.2.0
-appVersion: v1.21.1
+version: 1.22.1
```
ilyam8, line 4: Version is a semver of the helmchart, it wasnt equal to the netdata version before

Ferroin: @prologic Looks like there's a bug in the update script.

ilyam8:
@prologic https://github.com/netdata/helmchart/pull/98#discussion_r423994078

PR #102, opened 2020-06-04
Add service discovery sidecar to the nd slave

ilyam8:
Fixes: netdata/netdata#9021
Fixes: netdata/netdata#9015
Fixes: netdata/netdata#8925

blockers:
 - [x] netdata v1.23.0 with new go.d.plugin
 - [x] sd tagged version (netdata/agent-sd)
 - [x] enable cgroups plugin (disabled to remove noise in demo)
 - [x] sd default config

---

This PR:

- adds sd sidecar to the netdata slave pod definition
- adds sd configuration file (`sd-slave.yml`)
- sd is enabled by default (configurable option)

sd <-> netdata (go.d.plugin) communicate via configuration file (sd produces, go.d consumes), so a new shared dir (type emptyDir) is added too.

---

Tests:
 - [x] check templates generation, no errors (`helm template .`)
 - [x] install this branch on my test gke cluster, ensure both `sd` and `netdata` is up and running and there are no errors
 - [x] ensure `sd` sidecar is not added if it is disabled in the values


ilyam8:
btw i dont get why all these rules are in cluster role

```yaml
rules:
  - apiGroups: [""]
    resources:
      - "nodes"
      - "pods"
      - "services"
      - "endpoints"
      - "configmaps"
      - "secrets"
      - "events"
      - "componentstatuses"
      - "nodes/proxy"
    verbs:
      - "get"
      - "list"
      - "watch"
  - apiGroups: [""]
    resources:
      - "resourcequotas"
    verbs:
      - "get"
      - "list"
  - apiGroups: [""]
    resources:
      - "nodes/metrics"
      - "nodes/spec"
    verbs:
      - "get"
  - apiGroups: ["extensions"]
    resources:
      - "ingresses"
    verbs:
      - "get"
      - "list"
      - "watch"
  - nonResourceURLs: ["/version", "/healthz", "/metrics"]
    verbs:
      - "get"
```

current sd needs only

```yaml
rules:
  - apiGroups: [""]
    resources:
      - "pods"
      - "services"
      - "configmaps"
      - "secrets"
    verbs:
      - "list"
      - "watch"
```

In future it is likely we will need endpoints + nodes + ingresses.

I think it is better to have only those we need. Because it is confusing atm.

ilyam8:
@prologic do you have admin access to the repo?

```
Settings => 
Branches => 
Branch protection rules => 
Edit => 
Uncheck "Dismiss stale pull request approvals when new commits are pushed"
```

templates/configmap.yaml
```diff
@@ -1,35 +1,53 @@
...
+metadata:
+  name: netdata-slave-sd-config-map
+  labels:
+    app: {{ template "netdata.name" . }}
+    chart: {{ template "netdata.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+data:
+  config.yml: |
+  {{- .Files.Get .Values.sd.slave.file | nindent 4 }}
```
ncmans, line 52: How can the chart user override this without modifying the chart?

ilyam8: overwrite what?

file is configurable

https://github.com/ilyam8/helmchart/blob/b41183085a7c09fdde46bffefbc40074f2287f35/values.yaml#L14

ilyam8: if you mean the file - copy paste the file, customize it to fit his need. It looks better then have all the configuration in the `values.yml` - there is only file name to use in the values.

ilyam8: [i did it like that at first](https://github.com/netdata/helmchart/pull/102/commits/b19c21fc7e39cdbda6f2993e1eebea4a50761223) but it makes `values.yml` really big, having the config in a separate file serve same purpose and looks more clear to me.

ncmans: > if you mean the file - copy paste the file, customize it to fit his need

Common practice in cases like this is to allow the user to override the configmap name.

ilyam8: can you elaborate how it works from your perspective?

i see we dont override config map names
> https://github.com/netdata/helmchart/blob/master/templates/configmap.yaml

configuration is a file, use can have its own configuration and override the  filename option. Why it doesnt do?

ilyam8: like

```yaml
sd:
  repository: netdata/agent-sd
  tag: latest
  pullPolicy: Always
  slave:
    enabled: true
    file: sd-slave.yml
    configmap:
      name: <name>
      key: <name>
    resources: {}
```

And if `configmap.name` and `configmap.key` *are set* we dont generate our configmap and use custom user cmap?

ilyam8: i was thinking about something like that, i decided not to do it because - it is easy to add options, it is hard to remove options, i decided to go with minimal config and if doesnt work for someone and there is a request to change/add something *only after it*, if reasoning is good - add it.

ncmans: > configuration is a file

User of helm shouldn't be expected to have a config file on their filesystem. For example, when using helm from terraform, it's almost never the case that a config file is first generated, saved to disk, and then referenced from a chart. More commonly, content of a file is passed to helm as value.

> a user can have its own configuration and override the filename option

I've never seen this done anywhere else and (not having actually tested it) don't even know if it works when installing a chart from a repository. Have you tried it?

> And if configmap.name and configmap.key are set we dont generate our configmap and use custom user cmap?

Yes exactly! But without the `file` option, as it no longer something the user cares about. As mentioned earlier, helm shouldn't rely on having access to filesystem of the user for values.

ncmans: > don't even know if it works when installing a chart

From [Helm Docs](https://helm.sh/docs/chart_template_guide/accessing_files/#lines):

> There is no way to pass files external to the chart during `helm install`

ilyam8: All the templates are on the filesystem, there is just another file.

> Have you tried it?

There is documentation

>  But sometimes it is desirable to import a file that is not a template and inject its contents without sending the contents through the template renderer.

> https://helm.sh/docs/chart_template_guide/accessing_files/

You can check it works (test i did in [the first post](https://github.com/netdata/helmchart/pull/102#issue-427930112))

> helm template -s templates/configmap.yaml .

---

Let me summarize, you suggestion is:
 - put default configuration to the `values.yml`
 - generate config map using this config, the name should be hardcoded
 - have an option to set configmap name/key
 - if both are set do not generate configmap with the config from values and  just use these parametes ([here](https://github.com/ilyam8/helmchart/blob/b41183085a7c09fdde46bffefbc40074f2287f35/templates/daemonset.yaml#L176))

?

ilyam8: >  installing a chart from a repository. Have you tried it?

i think i misunderstood you, what i did is `git clone/helm install`

ilyam8: ill think how to change the PR

ilyam8: will something like this do?

```yaml
sd:
  repository: netdata/agent-sd
  tag: latest
  pullPolicy: Always
  slave:
    enabled: true
    config: |
      ... data
    resources: {}
```

So a user can override just the whole config? Do we need cmap name/key configurable? Or you mean there is a chance the user might want to have this config in another configmap? not dedicated for the service?

ilyam8: it would be very nice you show the config example you have in mind if you have time

ncmans: We provide our chart from repository as well. We can't expect users to clone the helm chart, modify it (ie add a new file in the chart directory) and *then* install it, in order to override config.

> Let me summarize, your suggestion is:

Almost. Here's the slightly modified one:

* Keep the `sd-slave.yml` as is (to reap the benefits of clean `values.yaml`)
* Reference it in the template as you do currently.
* Remove the `sd.slave.file` from `values.yaml`
* Generate config map using this config, the name should be hardcoded
* Have an option to set configmap name/key
* If both are set do not generate configmap with the config from values and just use these parameters

ncmans: > it would be very nice you show the config example you have in mind if you have time

You had it down perfectly before:

```yaml
sd:
  repository: netdata/agent-sd
  tag: latest
  pullPolicy: Always
  slave:
    enabled: true
    configmap:
      name: <name>
      key: <name>
    resources: {}
```

ilyam8: i see you dont like `file` option, however not having it means i hardcode `sd-slave.yml` and there is no way to overwrite it (i think to configure a template is no-go), but if a had my own config and i use helm i would like to overwrite it, what if

```
sd:
  repository: netdata/agent-sd
  tag: latest
  pullPolicy: Always
  slave:
    enabled: true
    configmap:
      name: netdata-slave-sd-config-map
      key: config.yml
      file: sd-slave.yml
    resources: {}
```

if there is `file` we generate the map from file, if there is no file (or the value is `""`) we dont generate the map (in both cases we use name/key) ?

ilyam8: it could be even:

- generate from file
- or generate from a value
- do not generate and use just name/key


```yml
sd:
  repository: netdata/agent-sd
  tag: latest
  pullPolicy: Always
  slave:
    enabled: true
    configmap:
      name: netdata-slave-sd-config-map
      key: config.yml
      from:
        file: sd-slave.yml
        value: ""
    resources: {}
```

ilyam8: generating from a file (not hardcoded filename) needed at least by me

ilyam8: There could be different application identification patterns, we could have several configs in the repo and a user could just switch between them.

Datadog uses container name for app identification for example

> https://docs.datadoghq.com/agent/kubernetes/integrations/?tab=kubernetes#configuration

ncmans: sure, that sounds reasonable

ilyam8: thx for the help, i will do like that, i think it make sense to create a dir and put defaut config in there (better then have in the root)

ilyam8: @ncmans i think i addressed your suggestions

prologic:
> @prologic do you have admin access to the repo?
> 
> ```
> Settings => 
> Branches => 
> Branch protection rules => 
> Edit => 
> Set "Dismiss stale pull request approvals when new commits are pushed"
> ```

Done 👍

PR #104, opened 2020-06-25
Bump 1.23.0

prologic:


Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 name: netdata
 home: https://github.com/netdata/netdata
-version: 1.3.0
-appVersion: v1.22.1
+version: 1.3.1
```
ilyam8, line 4: Why we update patch? Or it doesnt matter and the goal to increase the number and that is it?

prologic: Pretty much. There are no significant changes to the Chart itself so hence the patch level bump here.

ilyam8: well

> Helm uses a packaging format called charts. A chart is a collection of files that describe a related set of Kubernetes resources.

We added new functionality in https://github.com/netdata/helmchart/pull/102

I think should be 1.4.0

> https://semver.org/#summary

PR #120, opened 2020-07-28
Convert parent statefulset to deployment

### LINKED ISSUE #86: netdata master: survive node failure
Netdata master runs as `Statefulset`.

> https://kubernetes.io/docs/tasks/run-application/force-delete-stateful-set-pod/#statefulset-considerations

> StatefulSet ensures that, at any time, there is at most one Pod with a given identity running in a cluster.

It means that if the node with Netdata master is not available (network problems, power problems, etc) statefulset woudlnt be rescheduled on the other node of the cluster. Which bascially means no monitoring.

Related issues:
 - https://github.com/kubernetes/kubernetes/issues/54368
 - https://github.com/kubernetes/kubernetes/issues/74947

Suggested way to handle such kind of situation is manually delete the node from the cluster to get rescheduling statefulsets possible.

___

So, what is the best way to handle node with netdata master failure?

- Do not schedule netdata master on the worker node?
- Install netdata master outside of the cluster?

@knatsakis @ncmans

ncmans:
Fixes #86 

Tested locally on a kind cluster:

- PVC are created and mounted correctly
- Changing parent's deployment results in first the existing parent pod terminating and then a new one being created.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -1,4 +1,6 @@
 replicaCount: 1
+deploymentStrategy:
+  type: Recreate
 
 image:
   repository: netdata/netdata
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ This chart bootstraps a [Netdata](https://github.com/netdata/netdata) deployment
 cluster using the [Helm](https://helm.sh) package manager.
 
 The chart installs a Netdata child pod on each node of a cluster, using a `Daemonset` if not disabled, and a Netdata
-parent pod on one node, using a `Statefulset`. The child pods function as headless collectors that collect and forward
+parent pod on one node, using a `Deployment`. The child pods function as headless collectors that collect and forward
 all the metrics to the parent pod. The parent pod uses persistent volumes to store metrics and alarms, handle alarm
 notifications, and provide the Netdata UI to view metrics using an ingress controller.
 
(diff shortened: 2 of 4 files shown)


ilyam8:
@ncmans readme should be updated too

```cmd
0 helmchart (master)$ grep -i "statefulset" -R *
README.md:parent pod on one node, using a `Statefulset`. The child pods function as headless collectors that collect and forward
README.md:`replicaCount` | Number of `replicas` for the parent netdata `Statefulset` | `1`
README.md:`parent.resources` | Resources for the parent statefulset | `{}`
README.md:`parent.nodeSelector` | Node selector for the parent statefulset | `{}`
README.md:`parent.tolerations` | Tolerations settings for the parent statefulset | `[]`
README.md:`parent.affinity` | Affinity settings for the parent statefulset | `{}`
README.md:`parent.priorityClassName` | Pod priority class name for the parent statefulset | `""`
README.md:`parent.env` | Set environment parameters for the parent statefulset | `{}`
templates/statefulset.yaml:kind: StatefulSet
```

values.yaml
```diff
@@ -1,4 +1,6 @@
 replicaCount: 1
```
ilyam8, line 1: i see zero sense in `replicaCount`, am i missing something?

ncmans: That wasn't added in this PR. It's however common practice to keep it for folks who wish to do testing. One such use is to set it to zero to effectively disable the parent.

PR #129, opened 2020-08-13
Fix storageClass condition

srueg:
The condition for the database storageClass is wrong (copy paste error?).


ncmans:
Yes, thank you.

PR #134, opened 2020-09-08
sdconfig: add prometheus endpoints identification using `prometheus.io/scrape` annotation

ilyam8:
blocked by:
- [x] netdata v1.25.0 release
- [x] sd new version release

Blockers are resloved.

---

Fixes: netdata/netdata#9774

This PR changes service discovery config, it adds identification prometheus enpoints via well-known annotations:

- `prometheus.io/scrape`: The default configuration will scrape all pods and, if set to false, this annotation will exclude the pod from the scraping process.
- `prometheus.io/path`: If the metrics path is not /metrics, define it with this annotation.
- `prometheus.io/port`: Scrape the pod on the indicated port instead of the pod’s declared ports (default is a port-free target if none are declared).

I tested this PR locally in minikube and on gke.

### DIFF
--- a/sdconfig/child.yml
+++ b/sdconfig/child.yml
@@ -69,7 +69,7 @@ tag:
         expr: |
           {{ $scrapeOK := eq (get .Annotations "prometheus.io/scrape") "true" -}}
           {{ $portOK := eq (default .Port (get .Annotations "prometheus.io/port")) .Port -}}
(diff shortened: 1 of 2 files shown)


prologic (approved):

ncmans (approved):
LGTM as far as it installs and doesn't seem to have any obvious issues.

Issue #140, opened 2020-09-18
helm chart fails using argocd


MansM:
```
KIND                NAMESPACE NAME                                           STATUS.      MESSAGE
v1/ConfigMap netdata.          netdata-child-sd-config-map.  SyncFailed. error validating data: ValidationError(ConfigMap): unknown field "---apiVersion" in io.k8s.api.core.v1.ConfigMap
```

my guestimate is that some if the "-" are rendered instead of being part of the logic...
https://github.com/netdata/helmchart/blob/master/templates/configmap.yaml

ilyam8:
There were no changes for a while in that file, it works


> helm template -s templates/configmap.yaml .


```yaml
---
# Source: netdata/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: netdata-conf-parent
  labels:
    app: netdata
    chart: netdata-2.0.8
    release: RELEASE-NAME
    heritage: Helm
data:
  health:     |
      SEND_EMAIL="NO"
      SEND_SLACK="YES"
      SLACK_WEBHOOK_URL=""
      DEFAULT_RECIPIENT_SLACK=""
      role_recipients_slack[sysadmin]="${DEFAULT_RECIPIENT_SLACK}"
      role_recipients_slack[domainadmin]="${DEFAULT_RECIPIENT_SLACK}"
      role_recipients_slack[dba]="${DEFAULT_RECIPIENT_SLACK}"
      role_recipients_slack[webmaster]="${DEFAULT_RECIPIENT_SLACK}"
      role_recipients_slack[proxyadmin]="${DEFAULT_RECIPIENT_SLACK}"
      role_recipients_slack[sitemgr]="${DEFAULT_RECIPIENT_SLACK}"
  netdata:     |
      [global]
        memory mode = save
        bind to = 0.0.0.0:19999
      [plugins]
        cgroups = no
        tc = no
        enable running new plugins = no
        check for new plugins every = 72000
        python.d = no
        charts.d = no
        go.d = no
        node.d = no
        apps = no
        proc = no
        idlejitter = no
        diskspace = no
  stream:     |
      [11111111-2222-3333-4444-555555555555]
        enabled = yes
        history = 3600
        default memory mode = save
        health enabled by default = auto
        allow from = *
---
# Source: netdata/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: netdata-conf-child
  labels:
    app: netdata
[...]

PR #141, opened 2020-09-19
Update configmap.yaml

MansM:
fixes: https://github.com/netdata/helmchart/issues/140

tested with helm 2 and 3

### DIFF
--- a/templates/configmap.yaml
+++ b/templates/configmap.yaml
@@ -17,7 +17,7 @@ data:
 
 ---
 
-{{ if .Values.child.enabled }}
+{{- if .Values.child.enabled }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
@@ -37,7 +37,7 @@ data:
 
 ---
 
-{{ if and .Values.child.enabled .Values.sd.child.enabled .Values.sd.child.configmap.from }}
+{{- if and .Values.child.enabled .Values.sd.child.enabled .Values.sd.child.configmap.from }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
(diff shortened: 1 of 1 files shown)


templates/configmap.yaml
```diff
@@ -37,7 +37,7 @@ data:
 
 ---
 
-{{- if and .Values.child.enabled .Values.sd.child.enabled .Values.sd.child.configmap.from -}}
+{{ if and .Values.child.enabled .Values.sd.child.enabled .Values.sd.child.configmap.from }}
```
ilyam8, line 40: This works, but i think we need to remove only the right delimiter `-`

templates/configmap.yaml
```diff
@@ -17,7 +17,7 @@ data:
 
 ---
 
-{{- if .Values.child.enabled -}}
+{{ if .Values.child.enabled }}
```
ilyam8, line 20: same

templates/configmap.yaml
ilyam8: otherwise there is extra empty line

ncmans (approved):

knatsakis (approved):

prologic (changes requested):
LGTM but @ilyam8 is [right](https://pkg.go.dev/text/template#hdr-Text_and_spaces) Pleaseamend.

MansM:
I am sorry, I am still down with a nasty fever and didnt had time to get to it earlier

ilyam8 (approved):

prologic (approved):

knatsakis:
Thank you for your contribution @MansM

PR #142, opened 2020-09-20
Add ability to control graceful shutdown period

kam1kaze:
The default shutdown time (30 seconds) is not enough for parent node in our case. Time-to-time we get a corrupted database during node maintenance/lifecycle (we are using preemptible nodes on dev environment). So we would like to control shutdown time.

BTW, lifecycle.preStop does not help in this case. Bcz if the preStop hook hangs during execution, the Pod phase stays in a Terminating state and is killed after terminationGracePeriodSeconds of pod ends [(doc)](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#hook-handler-execution).

I would like to suggest to increase the default value of `parent.terminationGracePeriodSeconds` up to 300 seconds by default. Such value will be more friendly for users with slow IO or big databases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,7 @@ Parameter | Description | Default
 `clusterrole.name` | Name of the cluster role linked with the service account | `netdata`
 `APIKEY` | The key shared between the parent and the child netdata for streaming | `11111111-2222-3333-4444-555555555555`
 `parent.resources` | Resources for the parent deployment | `{}`
+`parent.terminationGracePeriodSeconds` | Duration in seconds the pod needs to terminate gracefully | `300`
 `parent.nodeSelector` | Node selector for the parent deployment | `{}`
 `parent.tolerations` | Tolerations settings for the parent deployment | `[]`
 `parent.affinity` | Affinity settings for the parent deployment | `{}`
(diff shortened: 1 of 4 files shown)


ncmans:
Thanks @kam1kaze for your contribution. LGTM.

ncmans (approved):

knatsakis (approved):

prologic (approved):

knatsakis:
> I would like to suggest to increase the default value of `parent.terminationGracePeriodSeconds` up to 300 seconds by default. Such value will be more friendly for users with slow IO or big databases.

I am happy with this. If others are happy too, do you mind amending your PR at setting the default to 300? Thanks!

prologic:
> I am happy with this. If others are happy too, do you mind amending your PR at setting the default to 300? Thanks!

I'm fine with this also. My only concerns is that whatever default value we provide will never be good enough for every infrastructure depending on size and no. of children agents streaming to their parent. Feels like something i missing here but I just can't put my finger on it 🤔

kam1kaze:
Increased `parent.terminationGracePeriodSeconds` to 300 seconds

prologic (approved):

PR #147, opened 2020-09-25
Make image.tag in values.yaml follow Chart.AppVersion

knatsakis:
One of the advantages is that we need to update the version number in one place only.


ncmans (approved):

prologic (approved):

srueg:
You unfortunately missed adding the `tpl` function to the child DaemonSet: https://github.com/netdata/helmchart/blob/master/templates/daemonset.yaml#L91

PR #152, opened 2020-09-29
Add a official helm chart repo landing page

prologic:
- Closes #74 

Screen Shot 2020-08-12 at 15 35 34

PR #153, opened 2020-09-29
Add Github Actions Workflows to automate Helm Chart Releases

### LINKED ISSUE #74: Make GitHub repo directly useable from Helm
Hi,
Until this is added to the official helm packages, it might be nice to have the helm Chart useable via this git repo as a helm repo?
I would like to install this using my terraform helm provider.
e.g. this should exist: https://raw.githubusercontent.com/netdata/helmchart/master/index.yaml

Ref https://github.com/netdata/helmchart/issues/72

prologic:
This now effectively gives us an official Helm Chart Repo.

Along with #152 we also have a nice looking landing page for it too!

Closes #74


.github/workflows/release.yml
```diff
@@ -0,0 +1,24 @@
...
+  push:
+    branches:
+      - master
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
```
kam1kaze, line 14: `actions/checkout` action fetches only last commit, but `helm/chart-releaser-action` uses `git diff` to determinate what HELM chart was changed and should be released.

So we have to add `fetch-depth: 0` param to fetch al git history, like in example here:
https://github.com/helm/chart-releaser-action#user-content-example-workflow

```
      - name: Checkout
        uses: actions/checkout@v2
        with:
          fetch-depth: 0
```

kam1kaze (changes requested):

knatsakis (dismissed):
LGTM as long as @kam1kaze's comment gets addressed

.github/workflows/release.yml
prologic: Done!

knatsakis (dismissed):

ncmans (approved):

prologic:
Rebased and Merging with prior approvals.

PR #154, opened 2020-09-29
templates/deployment.yaml: Run netdata as pid 1 also when using claiming

knatsakis:


knatsakis:
The docker image already does this https://github.com/netdata/netdata/blob/master/packaging/docker/run.sh#L22 when not using claiming.

kaskavel (approved):

ncmans (approved):

Issue #121, opened 2020-07-30
Child port exposure doesn't work as intended


P1ng-W1n:
According to current HELM chart, Netdata childs are exposed over the {{ .Values.service.port }} value:

From https://github.com/netdata/helmchart/blob/master/templates/daemonset.yaml#L118
```
ports:
            - name: http
              containerPort: {{ .Values.service.port }}
              protocol: TCP
```

However, that service port is intended to be used by the parent only (according to https://github.com/netdata/helmchart#configuration). If this value is changed, the container exported port will be different than the port netdata is actually bind to, as the port binding as affected by the ${NETDATA_PORT} environment variable, configured by the run.sh inside the container image:

https://github.com/netdata/netdata/blob/master/packaging/docker/Dockerfile#L109
```
ENTRYPOINT ["/usr/sbin/run.sh"]
```
https://github.com/netdata/netdata/blob/master/packaging/docker/run.sh#L22
```
exec /usr/sbin/netdata -u "${DOCKER_USR}" -D -s /host -p "${NETDATA_PORT}" -W set web "web files group" root -W set web "web files owner" root "$@"
```

As this environment variable is not passed to the child container as part of the HELM chart, so netdata will try to bind to the hardcoded 19999 port all the time.

Suggestions: 
-  Create a separate value in the values.yaml for child port exposure (as the current value is intended to be used by the parent only)
- Include the NETDATA_PORT variable as part of the HELM chart

ilyam8:
> Include the NETDATA_PORT variable as part of the HELM chart

I see we have `env` for both parent and child

https://github.com/netdata/helmchart/blob/a0b4cce6fd050558b23d2acb623cb5d162428d5b/values.yaml#L84-L86

https://github.com/netdata/helmchart/blob/a0b4cce6fd050558b23d2acb623cb5d162428d5b/values.yaml#L256-L257

So `NETDATA_PORT` can be set already

---

About `containerPort`, we could use something like (didnt test it)


```yaml
containerPort: {{ default .Values.service.port (int64 .Values.child.env.NETDATA_PORT) }}
```

---

But i see we have some hardcoded stuff

https://github.com/netdata/helmchart/blob/a0b4cce6fd050558b23d2acb623cb5d162428d5b/values.yaml#L118-L121

https://github.com/netdata/helmchart/blob/a0b4cce6fd050558b23d2acb623cb5d162428d5b/values.yaml#L214-L217

P1ng-W1n:
[...]

PR #170, opened 2020-10-15
Children's streaming destination port is not hardcoded to 19999

knatsakis:
Part of #121


ncmans (approved):

Ferroin (approved):

ilyam8:
Just in case it wasn't noticed - release action fails because we dont change https://github.com/netdata/helmchart/blob/4705f76f8458c04cb74dc1a5d9aae224f69d37fe/charts/netdata/Chart.yaml#L4

knatsakis:
> Just in case it wasn't noticed - release action fails because we dont change
> 
> https://github.com/netdata/helmchart/blob/4705f76f8458c04cb74dc1a5d9aae224f69d37fe/charts/netdata/Chart.yaml#L4

That's ok. Thanks for pointing it out!

PR #172, opened 2020-10-20
Chart.yaml: General makeover, many small improvements and updates

knatsakis:


ilyam8 (approved):

ncmans (dismissed):

charts/netdata/Chart.yaml
```diff
@@ -1,15 +1,28 @@
-apiVersion: v1
+apiVersion: v2
```
ncmans, line 1: @knatsakis have you considered that this may break people using helm < 3.0 ? I'm not sure it will, but thought I'd mention it.

knatsakis: I have reverted `apiVersion` to v1.

Let upgrade to v2 after EOL of v1 this November: https://helm.sh/blog/helm-v2-deprecation-timeline/

knatsakis:
Merging based on prior approvals.

charts/netdata/Chart.yaml
knatsakis: Opened #173 for this

PR #184, opened 2020-11-26
Support claiming of children

knatsakis:

PR #185, opened 2020-12-03
Allow to set service annotations and externalTrafficPolicy

dex4er:
Hello,

I wanted to expose netdata's service with internal address on separate reserved IP. It is possible on GKE if I use additional annotations. My example setup is:

```yaml
config:
  service:
    annotations:
      cloud.google.com/load-balancer-type: internal
      networking.gke.io/internal-load-balancer-allow-global-access: "true"
    externalTrafficPolicy: Local
    loadBalancerIP: 10.40.1.2
    port: 80
    type: LoadBalancer
```

This patch tweaks services.yaml and allows to set separate port for service and for pods, set externalTrafficPolicy and annotations.

Additionaly some extra settings might be used: healthCheckNodePort and clusterIP.

These additional settings are enabled only for correct service types.


ilyam8:
Hi @dex4er, thx for the contribution! (and sorry for such a delay in reviewing it 😅 )

Pleae bump version of of the chart ([see the example commit](https://github.com/netdata/helmchart/pull/184/commits/e4f6b515148b938b8d1cb5af501e73b46fbb0772))

dex4er:
> Pleae bump version of the chart ([see the example commit](https://github.com/netdata/helmchart/pull/184/commits/e4f6b515148b938b8d1cb5af501e73b46fbb0772))

Hi @ilyam8 . I rebased PR and bumped version. Thanks!

ilyam8 (approved):

dex4er:
@Ferroin, @knatsakis, can you review it, please?

I'm afraid that my PR will require another rebase soon because version won't match again. Maybe this workflow when *I* update version is wrong? In my opinion the version number should be bumped after merge, not in the same PR. In best case it should be changed after review, not before. It is really hard to guess what will be a version in a day of merge and because of version mismatch PR is shown usually as "can't merge".

dex4er:
Rebased again.

@Ferroin , @knatsakis , could you review it, please?

knatsakis (approved):

knatsakis:
Hi @dex4er, thanks a lot for your PR and patience.

> Maybe this workflow when I update version is wrong? In my opinion the version number should be bumped after merge, not in the same PR. In best case it should be changed after review, not before. It is really hard to guess what will be a version in a day of merge and because of version mismatch PR is shown usually as "can't merge"

This is indeed an area that needs improvement, and we plan to improve time permitting.

PR #193, opened 2021-02-17
persist netdata's child /var/lib/netdata on the host node

ilyam8:
### Goal

To get Netdata Agent instances claiming properly working we need to persist across containers restarts/re-creation:

- `/var/lib/netdata/cloud.d/`
- `/var/lib/netdata/registry/netdata.public.unique.id`

Data collection instances are running as [DaemonSet](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/), i think it makes a lot of sense to use hosts nodes filesystem to store the data.

### Solution

[hostPath](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath) volume type is what we need. It allows to create (if it doesn't exist) a directory on the host system.

However, there are some limitations:

> The files or directories created on the underlying hosts are only writable by root. You either need to run your process as root in a privileged Container or modify the file permissions on the host to be able to write to a hostPath volume.

So there are several options:
 1. run your process as root in a privileged Container.
 2. modify the file permissions on the host to be able to write to a hostPath volume.

Both will do, but **2nd is preferable**. No need to run netdata process as root, not need to make the container privileged.

**So we need only to modify the dir permissions using initContainer functionality.**

### Implementation

This PR:

- removes `child.persistUniqueID` option 
- adds `child.persistence`

```yaml
child:
  persistence:
    enabled: true                        # whether or not to presist `/var/lib/netdata` in the `hostPath`
    hostPath: /var/lib/k8s-netdata-child # host node directory for storing data.
```

Default install:

```cmd
root@minikube:/opt# tree k8s-netdata/
k8s-netdata/
`-- var
    `-- lib
        `-- netdata
            |-- cloud.d
            |-- dbengine_multihost_size
            |-- god-jobs-statuses.json
            |-- lock
            |   |-- coredns.collector.lock
            |   |-- k8s_kubelet.collector.lock
            |   `-- sensors.collector.lock
            |-- netdata.api.key
            |-- pythond-jobs-statuses.json
            `-- registry
                `-- netdata.public.unique.id
```
 
### Warning

⚠️ After this update netdata parent will lose its children metrics due to new children `netdata.public.unique.id`. (only once)

---

This PR is tested locally in minikube and in the GKE. Works good.


Ferroin:
Agreed that the path for this should be configurable, and default to something that will not clash with any host-side install of Netdata.

Also agreed on the UID, we _really_ need to get to a point where we do not have that hard-coded anywhere...

ilyam8:
Thinking we would want to persist something else in the future (e.g. /etc/netdata). Also, in my opinion, it would be nice to store all that data in one dir (not have `/var/lib/k8s-netdata`, `/etc/k8s-netdata`)

Perhaps implement it like that:

```yaml
  persistence:
    rootDir: /opt/k8s-netdata
    netdataUserUID: 201
    varlib:
      enabled: true
```

@knatsakis @Ferroin

ilyam8:
@Ferroin 

- `/opt` doesn't work in gke.

> Warning  FailedMount  7s (x7 over 39s)  kubelet            MountVolume.SetUp failed for volume "persistencerootdir" : mkdir /opt/k8s-netdata: read-only file system

- `var/lib` works in gke. 

```cmd
ilya@gke-standard-cluster-cost-optimized-p-0d6216c8-ygw3 ~ $ ls -l /var/lib/k8s-netdata/var/lib/netdata/
total 64
drwxr-x--- 2 messagebus messagebus  4096 Feb 17 18:46 cloud.d
-rw-r--r-- 1 root       root           3 Feb 17 18:46 dbengine_multihost_size
-rw-rw---- 1 messagebus messagebus 44569 Feb 17 18:49 god-jobs-statuses.json
drwxr-x--- 2 messagebus messagebus  4096 Feb 17 18:49 lock
-rw-rwx--- 1 messagebus messagebus    36 Feb 17 18:46 netdata.api.key
drwxrwx--- 2 messagebus messagebus  4096 Feb 17 18:46 registry
```

---

Which directory is suitable for storage netdata's various data? I feel like `var/lib` is not the best. Perhaps something in `/usr`?

charts/netdata/values.yaml
```diff
@@ -241,7 +241,10 @@ child:
 
   podAnnotations: {}
 
-  persistUniqueID: true
+  persistence:
+    rootDir: /opt/k8s-netdata
```
ilyam8, line 245: Should find another default, `/opt` doesnt work in gke.

> MountVolume.SetUp failed for volume "persistencerootdir" : mkdir /opt/k8s-netdata: read-only file system

knatsakis (approved):

PR #194, opened 2021-02-17
Add a GHA workflow for automating agent version updates.

Ferroin:
This adds a manually triggered workflow for generating PRs to update the version of the Netdata Agent provided by the Helm chart.

Short-term, this will be manually triggered during the release process, and will simply serve to speed things up a bit.

Longer-term, this will be automatically triggered by the release automation code in the agent repo once the Docker images have been published, reducing the amount of manual involvement required from the release manager.

PR #196, opened 2021-02-23
Update to agent version 1.29.3.

Ferroin:
SSIA

### DIFF
--- a/charts/netdata/Chart.yaml
+++ b/charts/netdata/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: netdata
-version: 3.6.2
+version: 3.6.3
 description: Real-time performance monitoring, done right!
 type: application
 keywords:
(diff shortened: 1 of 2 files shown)


charts/netdata/Chart.yaml
```diff
@@ -1,28 +1,28 @@
...
-    email: knatsakis@netdata.cloud
-  - name: Mansour Behabadi
-    email: mansour@netdata.cloud
-  - name: Vladimir Ryumin
-    email: vryumin@gmail.com
-  - name: Karuppiah Natarajan
-    email: karuppiah7890@gmail.com
-  - name: Oleksii Kravchenko
-    email: kamikaze.ua@gmail.com
+- name: Ilya Mashchenko
```
ilyam8, line 15: why not indent sequence value?

Ferroin: I was using the script from #194 and forgot to fix the indentation.

I still need to test if the script works as we need it to with ruamel.

ilyam8 (approved):

knatsakis (approved):

PR #205, opened 2021-04-22
Autodetect apiVersion for Ingress

dex4er:
[kubent](https://github.com/doitintl/kube-no-trouble) shows me deprecations for netdata's Ingress resource.

The patch implements autodetection of apiVersion for this resource. It is the same method as in bitnami's helm charts.


ilyam8:
Hi @dex4er 

Thanks for the contribution!

Related issue/PR https://github.com/netdata/helmchart/pull/67

---

>  It is the same method as in bitnami's helm charts.

Could you provide a link?

The approach is nice.

charts/netdata/templates/_helpers.tpl
```diff
@@ -30,3 +30,33 @@ Create chart name and version as used by the chart label.
...
+{{- print "networking.k8s.io/v1" -}}
+{{- end }}
+{{- else if semverCompare "<1.14-0" (include "netdata.kubeVersion" .) -}}
+{{- print "extensions/v1beta1" -}}
+{{- else if semverCompare "<1.19-0" (include "netdata.kubeVersion" .) -}}
+{{- print "networking.k8s.io/v1beta1" -}}
+{{- else -}}
+{{- print "networking.k8s.io/v1" -}}
+{{- end -}}
+{{- end -}}
```
ilyam8, line 62: It looks like we can simplify it.

Correct me if i am wrong:
 - if `.Values.ingress.apiVersion` is set we use it
 - otherwise we auto-detect using `netdata.kubeVersion`

So it is

```go
{{- define "netdata.ingress.apiVersion" -}}
{{- if .Values.ingress.apiVersion -}}
{{- .Values.ingress.apiVersion -}}
{{- else if semverCompare "<1.14-0" (include "netdata.kubeVersion" .) -}}
{{- print "extensions/v1beta1" -}}
{{- else if semverCompare "<1.19-0" (include "netdata.kubeVersion" .) -}}
{{- print "networking.k8s.io/v1beta1" -}}
{{- else -}}
{{- print "networking.k8s.io/v1" -}}
{{- end -}}
{{- end -}}
```

ilyam8: > if .Values.ingress.apiVersion is set we use it

And why/when we need to manually set ingress apiVersion?

dex4er:
@ilyam8 Bitnami's routines are at https://github.com/bitnami/charts/blob/master/bitnami/common/templates/_capabilities.tpl

charts/netdata/templates/_helpers.tpl
dex4er: I think manual setting might be useful if you prepare "helm template" for applying it later with "kubectl apply".

dex4er: Ok i think we can simplyfy it the way you did.

ilyam8 (approved):

knatsakis (approved):

knatsakis:
@dex4er thank you very much for your contribution!

PR #207, opened 2021-05-04
Restructure agent PR generation to produce verified commits.

Ferroin:
This restructures the GHA workflow responsible for generating agent version update PRs so that it produces verified commits by using the GitHub API to generate the commit instead of generating it locally on the runner and then pushing to GitHub. Making sure these commits are verified is important to helping foster user trust.

The resultant workflow is technically more verbose, but actually has one less step than the original approach did.

### DIFF
--- a/.github/workflows/agent-pr.yml
+++ b/.github/workflows/agent-pr.yml
@@ -22,16 +22,25 @@ jobs:
           git config --global user.name Netdatabot
           git config --global user.email bot@netdata.cloud
       - name: Create Branch
-        run: git checkout -b agent-${{ github.event.inputs.agent_version }}
+        run: |
+          git checkout -b agent-${{ github.event.inputs.agent_version }}
+          git push -u origin agent-${{ github.event.inputs.agent_version }}
       - name: Update Files
         run: .github/scripts/update-agent-version.py ${{ github.event.inputs.agent_version }}
       - name: Commit Changes
-        run: git commit -a -m 'Update agent version to ${{ github.event.inputs.agent_version }}.'
-      - name: Push Branch
-        run: git push -u origin agent-${{ github.event.inputs.agent_version }}
+        uses: swinton/commit@v2.x
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          files: |
+            charts/netdata/Chart.yaml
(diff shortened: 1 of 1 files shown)


.github/workflows/agent-pr.yml
```diff
@@ -22,16 +22,25 @@ jobs:
           git config --global user.name Netdatabot
           git config --global user.email bot@netdata.cloud
       - name: Create Branch
-        run: git checkout -b agent-${{ github.event.inputs.agent_version }}
+        run: |
+          git checkout -b agent-${{ github.event.inputs.agent_version }}
+          git push -u origin agent-${{ github.event.inputs.agent_version }}
```
ilyam8, line 27: Why do we need this during _Create Branch_?

Ferroin: The branch needs to exist on GitHub for the commit action later to actually work. Just pushing it like this is the quickest way to ensure that.

ilyam8 (approved):

ilyam8:
Btw [Update agent version to v1.30.1](https://github.com/netdata/helmchart/commit/7f32dada6c881122ecbab79ec19df398dea3cfa8) commit is verified

.github/workflows/agent-pr.yml
```diff
@@ -22,16 +22,25 @@ jobs:
...
+          files: |
+            charts/netdata/Chart.yaml
+            charts/netdata/README.md
+          commit-message: 'Update agent version to ${{ github.event.inputs.agent_version }}.'
+          ref: refs/heads/agent-${{ github.event.inputs.agent_version }}
       - name: Create PR
         uses: repo-sync/pull-request@v2
         with:
           source_branch: agent-${{ github.event.inputs.agent_version }}
           pr_title: 'Update agent version to ${{ github.event.inputs.agent_version }}.'
```
ilyam8, line 44:
```suggestion
          pr_title: 'Update agent version to ${{ github.event.inputs.agent_version }}'
```

neat: `Update agent version to v1.30.1.` => `Update agent version to v1.30.1` (https://github.com/netdata/helmchart/pull/203)

I think it looks better w/o `.`.

ilyam8 (approved):

knatsakis (approved):

PR #210, opened 2021-05-20
Fix unknown field "serviceName" in io.k8s.api.networking.v1.IngressBackend for Kubernetes 1.19+

cowbon:
`networking.k8s.io/v1alpha1` is deprecated for Kubernetes 1.21, causing `Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: [ValidationError(Ingress.spec.rules[0].http.paths[0].backend): unknown field "serviceName" in io.k8s.api.networking.v1.IngressBackend, ValidationError(Ingress.spec.rules[0].http.paths[0].backend): unknown field "servicePort" in io.k8s.api.networking.v1.IngressBackend]`. This PR updates the API definition to `networking.k8s.io/v1`

### DIFF
--- a/charts/netdata/templates/ingress.yaml
+++ b/charts/netdata/templates/ingress.yaml
@@ -26,8 +26,11 @@ spec:
       http:
         paths:
           - path: {{ $ingressPath }}
+            pathType: Prefix
             backend:
-              serviceName: {{ $fullName }}
-              servicePort: http
+              service:
+                name: {{$fullName}}
+                port:
+                  name: http
   {{- end }}
 {{- end }}


ilyam8:
Hi @cowbon 👋  


You are right, there is a bug. But your fix breaks compatibility with older versions.

`networking.k8s.io/v1beta1` is still valid in v1.21 ([removed in v1.22](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122)).

cc @dex4er

I guess a quick fix is to change

https://github.com/netdata/helmchart/blob/b4e5795ca95ecc6d3a39709d2d4121887fb06627/charts/netdata/templates/_helpers.tpl#L49-L52

to

```go
{{- else if semverCompare "<1.19-0" (include "netdata.kubeVersion" .) -}}
{{- "networking.k8s.io/v1beta1" -}}
{{- else -}}
{{- "networking.k8s.io/v1beta1" -}}
```

And do appropriate fix later.

---

Also, you can set the API version manually (.Values.ingress.apiVersion).

dex4er:
Hi @ilyam8 

I wouldn't stay with old API because it will cause deprecations warning in next K8S versions soon. I already send new PR #211 which should fix this issue.

I already tested my patch on K8S 1.18 and 1.19 and works fine in both cases.

ilyam8:
@cowbon closing this PR in favor of #211

PR #211, opened 2021-05-21
Fix Ingress for K8s >= 1.19

dex4er:
I'm really sorry that my #205 was not completed and I forgot to patch the Ingress object.

This PR should fix compatibility with K8S >= 1.19 and replaces #210

This is correct that K8S 1.19 still supports networking.k8s.io/v1beta1 but then very soon kubectl will report deprecations and it is better idea to detect used version and change slightly the resource based on desired version.

In Google Cloud now 1.19 is a main version since this week and then more and more people will need this bug fixed.

### DIFF
--- a/charts/netdata/templates/ingress.yaml
+++ b/charts/netdata/templates/ingress.yaml
@@ -1,8 +1,9 @@
 {{- if .Values.ingress.enabled -}}
 {{- $fullName := include "netdata.name" . -}}
 {{- $ingressPath := .Values.ingress.path -}}
+{{- $apiVersion := (include "netdata.ingress.apiVersion" .) -}}
 
-apiVersion: {{ include "netdata.ingress.apiVersion" . }}
+apiVersion: {{ $apiVersion }}
 kind: Ingress
 metadata:
   name: {{ $fullName }}
@@ -26,8 +27,18 @@ spec:
       http:
         paths:
           - path: {{ $ingressPath }}
+            {{- if and (ne $apiVersion "extensions/v1beta1") (ne $apiVersion "networking.k8s.io/v1beta1") }}
+            pathType: Prefix
+            {{- end }}
             backend:
+              {{- if or (eq $apiVersion "extensions/v1beta1") (eq $apiVersion "networking.k8s.io/v1beta1") }}
               serviceName: {{ $fullName }}
               servicePort: http
+              {{- else }}
+              service:
+                name: {{ $fullName }}
+                port:
+                  name: http
+              {{- end }}
   {{- end }}
 {{- end }}
--- a/charts/netdata/templates/ingress.yaml
+++ b/charts/netdata/templates/ingress.yaml
@@ -27,11 +27,11 @@ spec:
       http:
         paths:
           - path: {{ $ingressPath }}
-            {{- if and (ne $apiVersion "extensions/v1beta1") (ne $apiVersion "networking.k8s.io/v1beta1") }}
+            {{- if not (eq $apiVersion "extensions/v1beta1" "networking.k8s.io/v1beta1") }}
             pathType: Prefix
             {{- end }}
             backend:
-              {{- if or (eq $apiVersion "extensions/v1beta1") (eq $apiVersion "networking.k8s.io/v1beta1") }}
+              {{- if eq $apiVersion "extensions/v1beta1" "networking.k8s.io/v1beta1" }}
               serviceName: {{ $fullName }}
               servicePort: http
               {{- else }}
--- a/charts/netdata/Chart.yaml
+++ b/charts/netdata/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: netdata
-version: 3.6.9
+version: 3.6.10
 description: Real-time performance monitoring, done right!
 type: application
 keywords:


ilyam8:
> was not completed and I forgot to patch the Ingress object.

Yeah, I missed that too somehow, my IDE even shows it 😅 

<details>
<summary>screenshot [Click]</summary>

Screenshot 2021-05-21 at 12 03 35

</details>

charts/netdata/templates/ingress.yaml
```diff
@@ -26,8 +27,18 @@ spec:
       http:
         paths:
           - path: {{ $ingressPath }}
+            {{- if and (ne $apiVersion "extensions/v1beta1") (ne $apiVersion "networking.k8s.io/v1beta1") }}
```
ilyam8, line 30: perhaps make `isOldAPI` var and use it?

```
if not isOldAPI
 ...
else
 ...
```

At the moment we need not complex _ifs_ here, but only old/not old

dex4er: It might be more difficult to maintain because it might be networking.k8s.io/v2 someday and then you'll have 2 old APIs then more variables and then using explicit strings in condition is more descriptive I think.

ilyam8: > it might be networking.k8s.io/v2 someday

Or might not be. Your point is valid, but at the moment the condition is simple: is `networking.k8s.io/v1` or not.

When we have `networking.k8s.io/v2` we no longer support `extensions/v1beta1`, etc.

charts/netdata/templates/ingress.yaml
```diff
@@ -26,8 +27,18 @@ spec:
...
+            pathType: Prefix
+            {{- end }}
             backend:
+              {{- if or (eq $apiVersion "extensions/v1beta1") (eq $apiVersion "networking.k8s.io/v1beta1") }}
```
ilyam8, line 34:
```suggestion
              {{- if eq $apiVersion "extensions/v1beta1" "networking.k8s.io/v1beta1" }}
```

> eq (only) accepts two or more arguments and compares the second and subsequent to the first, returning in effect

charts/netdata/templates/ingress.yaml
ilyam8, line 30:
```suggestion
            {{- if not (eq $apiVersion "extensions/v1beta1" "networking.k8s.io/v1beta1") }}
```

> eq (only) accepts two or more arguments and compares the second and subsequent to the first, returning in effect

charts/netdata/templates/ingress.yaml
ilyam8: I hope the syntax is correct, didn't test. But you get the idea.

charts/netdata/templates/ingress.yaml
ilyam8: I hope the syntax is correct, didn't test. But you get the idea.

dex4er: No, unfortunately it doesn't work for me correctly. I'm afraid it has to be like that.

My intention was to make this Ingress resource compatible with new versions from the future. I can't just make simple condition `eq "networking.k8s.io/v1"` because future version most likely will use this field too. Then I need to make negative condition twice for `"extensions/v1beta1"` and `"networking.k8s.io/v1beta1"`.

Go Templaces syntax is a little inconvenient but at least works.

ilyam8: Sorry, what didn't work? What I did is simplified the expressing using `eq` function feature

> For simpler multi-way equality tests, eq (only) accepts two or more arguments and compares the second and subsequent to the first, returning in effect
> 
> arg1==arg2 || arg1==arg3 || arg1==arg4 ...
>
> (Unlike with || in Go, however, eq is a function call and all the arguments will be evaluated.)

=>


`and (ne $apiVersion "extensions/v1beta1") (ne $apiVersion "networking.k8s.io/v1beta1")`

same as

`not (eq $apiVersion "extensions/v1beta1" "networking.k8s.io/v1beta1")`

which you read as `if apiVersion is not any of ...`

What am i missing?

dex4er: I was missleaded by official documentation for Helm. https://helm.sh/docs/chart_template_guide/function_list/#string-functions doesn't show anything about more than 2 args for `eq`.

Variant with `{{- if not (eq $apiVersion "extensions/v1beta1" "networking.k8s.io/v1beta1") }}` works indeed.

ilyam8 (approved):

charts/netdata/templates/ingress.yaml
ilyam8: Oh, `eq` is [templates built-in](https://golang.org/pkg/text/template/). Helm extends available functions by using http://masterminds.github.io/sprig/

charts/netdata/templates/ingress.yaml
```diff
@@ -26,8 +27,18 @@ spec:
       http:
         paths:
           - path: {{ $ingressPath }}
+            {{- if not (eq $apiVersion "extensions/v1beta1" "networking.k8s.io/v1beta1") }}
+            pathType: Prefix
```
ilyam8, line 31: Actually, we can move `pathType` to the else branch like @chrisDeFouRire did in https://github.com/netdata/helmchart/pull/213/files

knatsakis (approved):

knatsakis:
Hi @dex4er ,

Thanks a lot for your contribution!

PR #212, opened 2021-05-26
use environment variables for claiming

ilyam8:
This PR changes claiming logic from using [command line](https://learn.netdata.cloud/docs/agent/claim#netdata-agent-command-line) to [environment variables](https://learn.netdata.cloud/docs/agent/claim#using-environment-variables).

The latter is better because we [don't try to claim on every container start if it is already claimed](https://github.com/netdata/netdata/blob/e4f0ae2290a36bb1b9184fffac1e9ed1f707c52e/packaging/docker/run.sh#L23-L28).

---

⚠️ Claiming using environment variables works since v1.30.0 ([there was a bug in prev versions](https://github.com/netdata/netdata/pull/10811)). Do we need to keep support for old versions? (fallback to _command line_ claiming if version < 1.30.0 ).

@Ferroin @knatsakis ?

**This PR doesn't keep compatibility: Netdata version < 1.30.0 + claiming doesn't work.**

---

Tests:

- install on GKE, ensure parent/children are claimed and connected to the Cloud.

---

~@Ferroin There is [_a bug_ in the netdata-claim.sh](https://github.com/netdata/netdata/blob/master/claim/netdata-claim.sh.in#L403-L422):~
 - ~we execute the script before starting the Agent~
 - ~it always `exit 5` the first time because the script expects the Agent to be running~

Fixed in https://github.com/netdata/netdata/pull/11195

### DIFF
--- a/charts/netdata/Chart.yaml
+++ b/charts/netdata/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: netdata
-version: 3.6.11
+version: 3.6.13
 description: Real-time performance monitoring, done right!
 type: application
 keywords:
--- a/charts/netdata/Chart.yaml
+++ b/charts/netdata/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: netdata
-version: 3.6.12
+version: 3.6.13
 description: Real-time performance monitoring, done right!
 type: application
 keywords:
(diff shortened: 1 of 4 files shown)


knatsakis (dismissed):
@ilyam8, thank you very much for this very nice PR!

Also, thanks go to @Ferroin for making it possible in the first place

ilyam8:
> why a user might want to use an old version of Netdata with the newest Helm chart though...

We've discussed it with @knatsakis and came to the same conclusion. I think we expect our users to use the latest stable Netdata with the Cloud.

knatsakis:
> We _probably_ should try to keep compatibility if possible. I can’t think of why a user might want to use an old version of Netdata with the newest Helm chart though...

The way I understand it, by seeing what other projects do with their Helm charts, is that a Helm chart revision is only meant to work with a specific docker image tag. These two go hand in hand. I don't think I have seen anyone doing it differently, except by chance.

I mean, a Helm chart revision /may/ work with an older/newer docker image tag but it is only because it happens to. Not because it was intended to.

ilyam8:
Supposed to be merged after #216 (because we manually set Chart version)

knatsakis (approved):

Ferroin (approved):

ilyam8:
We haven't released a new version since https://github.com/netdata/netdata/pull/11195. That means that non-zero RESTARTS (`kubectl get pods...`) are expected after deploying/claiming Netdata Agents first time.

PR #213, opened 2021-05-27
adds support for ingress v1 final api

chrisDeFouRire:
When trying to install netdata on a 1.20.2 cluster hosted on DigitalOcean DOKS, I encountered issues with the new v1 ingress API... Here's a fix to the helm chart to account for the new backend syntax.

I've only tested the fix on 1.20.2 but >=1.19.0 should work (and require the change).


chrisDeFouRire:
okay I should have checked the active pull requests in addition to the active issues... Sorry for the duplicate

Issue #165, opened 2020-10-14
Add readiness and liveness probes to child sd containers


ncmans:
From https://github.com/netdata/helmchart/issues/159

KonstantinosPapazis:
Not applicable.

PR #222, opened 2021-06-17
Kpap add liveness sd child

KonstantinosPapazis:
ref #165


ilyam8:
@KonstantinosPapazis Unfortunately, it won't do. [Service discovery](https://github.com/netdata/agent-service-discovery) has not REST API. It is a sidecar/one process container, I am not sure we need a readiness probe for it (nothing depends on the service-discovery). Same thoughts about liveness probe - if the process crashes the containers is being restarted.

> The kubelet uses liveness probes to know when to restart a container. For example, liveness probes could catch a deadlock, where an application is running, but unable to make progress.

I can't think of any such cases - sd app is running, but _unable to make progress_. I need to check it.

---

Should have commented in the issue 😅

Issue #81, opened 2020-04-01
add some checks/linters


ilyam8:
We need to add some yaml/helm linters.

ilyam8:
@ncmans could you recommend something?

ilyam8:
i guess https://github.com/helm/chart-testing-action is ok

prologic:
> i guess https://github.com/helm/chart-testing-action is ok

LGTM PR welcome :)

knatsakis:
Fixed in #227

PR #227, opened 2021-08-26
Adding checks Workflow

ShimonOhayon:
**Issue:** #81 
**Scope:**  Adding new workflow to check Helm charts in Pull Requests.

**Description:**

- Version bumping done via `Agent Version PR` Workflow, Which i could not test. I think we could enable version increment  checks here once you approve this. For now i kept it disable.

-  Helm insists that Chart Maintainers Names will be the VCS usernames,  I've tested it  the tests will pass if i will rename all maintainers names, as described [here](https://github.com/helm/chart-testing/issues/192). Can be enabled as well.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -1,0 +1,43 @@
+---
+name: Lint and Test Charts
+
+on: pull_request
+
+jobs:
+  lint-test:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


knatsakis (approved):

knatsakis:
Thank you @ShimonOhayon

Issue #229, opened 2021-09-14
Memory modes in parent and child should be changed


cakrit:
in values.yaml:

Parent config netdata should be changed to `memory mode = dbengine` 
It was originally set to `save` because the dbengine wasn't mature back in 2019. 

Child config netdata should be changed `memory mode = ram`. 
It was originally set to `none` because the parent is the only one we're supposed to be accessing via the UI. But since the cloud has for a while required child nodes to be claimed and we will continue to do this for high availability of the metrics, we should change the default to `ram`.

cakrit:
Resolved in PR #230

PR #230, opened 2021-09-14
changing default memory modes

M4itee:
ref #229
fixing the bug for issue mentioned https://community.netdata.cloud/t/metrics-missing-for-nodes-in-cloud/1697/5

### DIFF
--- a/charts/netdata/Chart.yaml
+++ b/charts/netdata/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: netdata
-version: 3.6.13
+version: 3.6.14
 description: Real-time performance monitoring, done right!
 type: application
 keywords:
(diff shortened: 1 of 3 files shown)


ilyam8:
Hey, @M4itee. We also need:
 - update [the README](https://github.com/netdata/helmchart#configuration-files) (`parent.configs.netdata` default value)
 - bump chart version ([we do it manually, yes](https://github.com/netdata/helmchart/pull/212/commits/0025d38235c172ea8ee294d40b6dcdabaa06c0a2))

ilyam8:
After this PR the parent metrics will be lost due to switching to the `dbengine` (nothing we can do about it). Also, users will need to adjust their configured retention policy (history => dbengine options).

charts/netdata/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: netdata
-version: 3.6.13
+version: 3.6.14
```
ilyam8, line 3: [Version badge](https://github.com/netdata/helmchart/tree/master/charts/netdata#netdata-helm-chart-for-kubernetes-deployments) needs to be updated too

ilyam8 (approved):

Issue #114, opened 2020-07-08
Clarify requirement for "alarms" PVC


cakrit:
The "alarms" directory has a lot more than alarms now (claiming information, machine guid and we'll be adding more). It's becoming really necessary to have this directory survive pod restarts.

We basically have three options here:
- Provide more options for persistence, that distinguish between the various files stored under `/var/lib/netdata`. A `claim.d` directory is getting added there as well. I'm not fond of this option, because it means we'll need to always add new options and require new volumes every time we add a new file or directory there.
- Change the docs and the config files so that it's clear it's not just `alarms` that supposedly use the persistent storage for `/var/lib/netdata` but also key things like the machine guid, the claiming information and the health management API key. We strongly recommend that users persist this volume and only disable it if they don't want to use netdata cloud.
- Just demand at least one persistent volume, to be used for `/var/lib/netdata`, removing completely the option to disable it.

Would like to hear your views.

ncmans:
I suggest:

- Having an optional single `data` persistent volume mounted under `/var/lib/netdata`.

prologic:
@joelhans As this is really documentation IHMO, Would you be able to take this on? The last two points are what I think we need here in terms of documentation improvements. Also @ncmans 's point of just having one persistent volume is right right approach here as well.

joelhans:
@prologic I can take a swing at this, sure. Once I see a PR for the volume, with the decision on whether to make it persistent or configurable, I'll follow up with an update to the docs.

ilyam8:
i think it should be configurable, a valid use case - install `nd` agent daemonset **only** w/o any streaming and cloud. People use such setups - store metrics in some tsdb and use `nd` agents for real-time debugging.

ncmans:
Yes, I believe a chunk of our users use the agent in a stateless config.

cakrit:
If we do that, the info will be useful only or mainly if they export the data to another persistent storage. I created #116 for it.

ncmans:
Given #116 is closed by a PR, is this issue resolved now?

PR #236, opened 2021-10-15
adding info about persistent volumes to the readme file

M4itee:
ref #114

### DIFF
--- a/charts/netdata/README.md
+++ b/charts/netdata/README.md
@@ -243,7 +243,7 @@ There are two different persistent volumes on `parent` node by design (not count
 1. database (`/var/cache/netdata`) - all metrics data is stored here. Performance of this volume affects query timings.
 2. alarms (`/var/lib/netdata`) - alarm log, if not persistent pod recreation will result in parent appearing as a new node in `netdata.cloud` (due to `./registry/` and `./cloud.d/` being removed).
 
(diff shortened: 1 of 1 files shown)


charts/netdata/README.md
```diff
@@ -237,6 +237,14 @@ must be indented with two more spaces relative to the preceding line:
         config line 2 #No problem indenting more here
 ```
 
+### Persistent volumes
+
+There are two different persistent volumes on `parent` node by design (not counting any Configmap/Secret mounts). Both can be used but they don't have to be. Keep in mind that whenever persistent volumes for `parent` are not used, all the data for specific PV is lost in case of pod removal.
+1. database (`/var/cache/netdata`) - all metrics data is stored here. Performance of this volume affects query timings.
+2. alarms (`/var/lib/netdata`) - alarm log, if not persistent pod recreation will result in parent appearing as a new node in `netdata.cloud` (due to `./registry/` and `./cloud.d/` being removed).
+
+In case of `childrens` it is a bit simpler. Most of the mounts we do share with the host, since by design it is a deamonset. By default hostPath: `/var/lib/netdata-k8s-child` is mounted on child in: `/var/lib/netdata`. Same rules apply here as in case of parrent's `alarms` volume.
```
ilyam8, line 246: For a child instance, it affects streaming. W/o that option, pod recreation results in appearing as a new replication node for the parent. I would say this option is kind of **required** unless you just want to test Netdata.

Issue #148, opened 2020-09-28
Is it a Security problem? very general PodSecurityPolicy


fzyzcjy:
Hi thanks for this helm chart! When reviewing the code, I find the [psp.yaml](https://github.com/netdata/helmchart/blob/master/templates/psp.yaml) allows k8s pods to do a lot of things (actually almost everything!). I am a new learner to k8s, so I wonder whether this is ok, or will it cause serious security problems? Thanks very much!

P.S. I know netdata needs some privileges, but it seems that the psp requires a lot, such as `allowedCapabilities: - '*'`. IMHO Netdata will only use a small portion among them.

ilyam8:
Related comment https://github.com/netdata/helmchart/pull/102#issuecomment-639086447

It would be nice if we have documented cluster role rules and psp.

fzyzcjy:
@ilyam8 Thanks very much! Seems that there should be fewer rules, so will it be updated?

ilyam8:
I think yes, we need to update it, what we have no is confusing.

fzyzcjy:
@ilyam8 Thanks and looking forward to that update!

ncmans:
We should only need:

- CAP_SYS_PTRACE
- Unconfined app-armour
- Passthrough `/proc`, `/sys`, `/`

Needs further investigation. We may be able to disable `privileged: true` in https://github.com/netdata/helmchart/blob/5fdfc74807c2a0ff707bb7ccd2ad0e425597e4b9/templates/deployment.yaml#L46

M4itee:
@ilyam8 could you please test the app with the settings in the PR #239 ?

M4itee:
due to #239  and #242 being merged - closing

PR #239, opened 2021-10-29
restrictions to psp

M4itee:
ref #148 
restrictions to the PSP:
- decrease in capabilities,
- privileged setting removed.

### DIFF
--- a/charts/netdata/templates/psp.yaml
+++ b/charts/netdata/templates/psp.yaml
@@ -9,8 +9,9 @@ metadata:
     chart: {{ .Chart.Name }}-{{ .Chart.Version }}
 spec:
   allowPrivilegeEscalation: true
-  defaultAddCapabilities:
+  allowedCapabilities:
     - SYS_PTRACE
+    - SYS_ADMIN
   fsGroup:
     rule: RunAsAny
   hostIPC: true
--- a/charts/netdata/templates/psp.yaml
+++ b/charts/netdata/templates/psp.yaml
@@ -20,6 +20,7 @@ spec:
   hostPorts:
   - max: 65535
     min: 0
+  privileged: true
   runAsUser:
     rule: RunAsAny
   seLinux:
--- a/charts/netdata/templates/psp.yaml
+++ b/charts/netdata/templates/psp.yaml
@@ -9,8 +9,8 @@ metadata:
     chart: {{ .Chart.Name }}-{{ .Chart.Version }}
 spec:
   allowPrivilegeEscalation: true
-  allowedCapabilities:
-  - '*'
+  defaultAddCapabilities:
+    - SYS_PTRACE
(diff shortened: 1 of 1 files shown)


knatsakis (approved):

rsawer:
did this pass functionality tests?

ilyam8 (changes requested):
Hold on with merging. I need to check it. I doubt that we can remove `privileged` from the sysctlInitContainer. That is the whole purpose of it - to be privileged and to be able to do changes on the host system (not sure tho).

M4itee:
@ilyam8 did you had a time to poke this issue with some kind of stick or not yet? :)

charts/netdata/templates/psp.yaml
```diff
@@ -9,8 +9,8 @@ metadata:
     chart: {{ .Chart.Name }}-{{ .Chart.Version }}
 spec:
   allowPrivilegeEscalation: true
-  allowedCapabilities:
-  - '*'
+  defaultAddCapabilities:
+    - SYS_PTRACE
```
ilyam8, line 13:
```suggestion
  allowedCapabilities:
    - SYS_PTRACE
    - CAP_SYS_ADMIN
```

Those are needed by [apps.plugin](https://github.com/netdata/netdata/tree/master/collectors/apps.plugin#appsplugin) and [cgroups.plugin](https://github.com/netdata/netdata/tree/master/collectors/cgroups.plugin#cgroupsplugin).

See https://github.com/netdata/helmchart/blob/915ea3bd202aef27385638dc73fc61071cecd40b/charts/netdata/templates/daemonset.yaml#L157-L161

And I think that 'allowedCapabilities' is better (more secure) than 'allowedCapabilities'.

charts/netdata/templates/psp.yaml
```diff
@@ -19,7 +19,6 @@ spec:
   hostPorts:
   - max: 65535
     min: 0
-  privileged: true
```
ilyam8, line 22: 'privileged: true' is needed for [sysctlInitContainer](https://github.com/netdata/helmchart/blob/915ea3bd202aef27385638dc73fc61071cecd40b/charts/netdata/templates/daemonset.yaml#L48-L56).

I don't know how people use this option (or is it used at all). My understanding is that sysctlInitContainer is supposed to perform changes on the host system thus needs to be priviliged.

ilyam8 (changes requested):

charts/netdata/templates/psp.yaml
ilyam8: Related article: https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/

ShimonOhayon (approved):

ilyam8 (approved):

PR #240, opened 2021-11-04
Automation of the Chart versions and releases

ShimonOhayon:
Ref: netdata/infra#1506
Changes in this PR:

- Refactoring the versioning script to decouple between agent (app) version and chart version
- Any push to master will trigger a Chart version bump and release, (For files changed under`charts/**` directory)
- The release will commit back the new version using admin user (`netdatabot`) to bypass branch protection.

With this changes we will have 1 release cycle only from 'master' branch that will contain both input for the chart version and app version. This will prevent block and conflicts.

Testing:
I've executed successfully the [agent-pr](https://github.com/ShimonOhayon/helmchart/actions/runs/1420481671) and [release](https://github.com/ShimonOhayon/helmchart/actions/runs/1418535826) workflows on the forked repo.


.github/scripts/update_versions.py
```diff
@@ -0,0 +1,105 @@
...
+            data = readme.read()
+            data = data.replace(str(old_version), str(new_version))
+            readme.seek(0)
+            readme.write(data)
+
+
+def main():
+    args = _parse_command_line()
+    chart_data = FileHandler.load_chart_file()
+    if args.action == 'set_app_version' and not args.version:
```
ShimonOhayon, line 99: forcing mutual inclusive args

.github/workflows/agent-pr.yml
```diff
@@ -26,15 +26,14 @@ jobs:
...
-        run: .github/scripts/update-agent-version.py ${{ github.event.inputs.agent_version }}
+        run: .github/scripts/update_versions.py set_app_version --version ${{ github.event.inputs.agent_version }}
       - name: Commit Changes
         uses: swinton/commit@v2.x
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           files: |
             charts/netdata/Chart.yaml
-            charts/netdata/README.md
```
ShimonOhayon, line 37: `README.md` will be update as part of the release process only to avoid pre mature update of the version before its actually been released

knatsakis (approved):
❤️ the use of `ruamel-yaml`

.github/workflows/release.yml
```diff
@@ -15,12 +17,29 @@ jobs:
...
 
       - name: Run chart-releaser
         uses: helm/chart-releaser-action@v1.2.1
         env:
           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+
+      - name: Commit new version
+        run: |
+          git config --global user.name Netdatabot
+          git config --global user.email bot@netdata.cloud
```
Ferroin, line 43: You shouldn’t need to set these again here, the config from the earlier job should still be in effect for the rest of the job.

ShimonOhayon: True, removing

knatsakis (approved):

.github/scripts/update_versions.py
```diff
@@ -0,0 +1,105 @@
...
+        print(self.chart_data['version'])
+
+    def set_chart_version(self):
+        old_chart_data = self.chart_data['version']
+        new_chart_version = semver.VersionInfo.parse(old_chart_data) \
+                                                     .next_version(part='patch')
+
+        FileHandler.write_readme(old_chart_data, new_chart_version)
+        self.chart_data['version'] = str(new_chart_version)
+        FileHandler.write_chart(self.chart_data)
```
ilyam8, line 63: This method doesn't update AppVersion in the README (which could be updated by agent-pr workflow)

.github/scripts/update_versions.py
```diff
@@ -0,0 +1,105 @@
...
+                '--version',
+                required=False,
+                type=str,
+                nargs='?',
+                help='Version to set.')
+    return parser.parse_args()
+
+
+class Versioning:
+    '''Getting current Chart and App versions and bumping to new ones.'''
```
ilyam8, line 44: nit: Triple double-quoted strings should be used for docstrings. 

```suggestion
    """Getting current Chart and App versions and bumping to new ones."""
```

.github/scripts/update_versions.py
```diff
@@ -0,0 +1,105 @@
...
+        print(self.chart_data['appVersion'])
+
+    def set_app_version(self):
+        new_app_version = self.version
+        self.chart_data['appVersion'] = new_app_version
+        FileHandler.write_chart(self.chart_data)
+
+
+class FileHandler(object):
+    ''' Handling the Chart and app versions in Chart.yaml and README.md files'''
```
ilyam8, line 75: nit: Triple double-quoted strings should be used for docstrings.

```suggestion
    """ Handling the Chart and app versions in Chart.yaml and README.md files"""
```

.github/scripts/update_versions.py
```diff
@@ -0,0 +1,105 @@
...
+
+def main():
+    args = _parse_command_line()
+    chart_data = FileHandler.load_chart_file()
+    if args.action == 'set_app_version' and not args.version:
+        print("You must specify an app version for the 'set_app_version' command")
+        sys.exit(1)
+    Versioning(chart_data, args.action, args.version).run()
+
+if __name__ == "__main__":
```
ilyam8, line 104: nit: 2 blank lines after class or function

```suggestion


if __name__ == "__main__":
```

.github/scripts/update_versions.py
ilyam8: I mean the badge

https://github.com/netdata/helmchart/blob/6dc7e8fcdfc8c58815f3de7332de9335cd721ac0/charts/netdata/README.md?plain=1#L3

ShimonOhayon: True i will add it

ilyam8:
@ShimonOhayon good work, thanks! I have a few suggestions:

 - remove `(e.g v1.31.0)` from https://github.com/netdata/helmchart/blob/6dc7e8fcdfc8c58815f3de7332de9335cd721ac0/charts/netdata/README.md?plain=1#L87, so we don't update this line. And I don't think that info is needed there ("latest stable image" is enough for description).
 - how about just updating "Artifact HUB" line when updating the readme? As for how it is now we can accidentally update something we don't want to be updated (e.g. chart/app version is the same as some other package version). What do you think?

knatsakis:
> @ShimonOhayon good work, thanks! I have a few suggestions:
> 
> * remove `(e.g v1.31.0)` from https://github.com/netdata/helmchart/blob/6dc7e8fcdfc8c58815f3de7332de9335cd721ac0/charts/netdata/README.md?plain=1#L87, so we don't update this line. And I don't think that info is needed there ("latest stable image" is enough for description).
> * how about just updating "Artifact HUB" line when updating the readme? As for how it is now we can accidentally update something we don't want to be updated (e.g. chart/app version is the same as some other package version). What do you think?

+1 for both suggestions

ShimonOhayon:
@ilyam8 Tanks for the input !
No problem to remove `(e.g v1.31.0)` from the table,

About the artifactHub Line update, Just to confirm you mean to keep the changes to this line to only the `version` and `appversion` badges will be updated and avoid search and replace on the entire file, right?

ilyam8:
@ShimonOhayon yes, correct.

.github/scripts/update_versions.py
```diff
@@ -0,0 +1,108 @@
...
+    @staticmethod
+    def write_chart(chart_data):
+        with open(PREFIX.joinpath('Chart.yaml'), 'w', encoding='UTF-8') as chart:
+            yaml.dump(chart_data, chart)
+
+    @staticmethod
+    def write_readme(old_version, new_version):
+        with open(PREFIX.joinpath('README.md'), 'r+', encoding='UTF-8') as readme:
+            data = readme.read()
+            data = data.replace(str(old_version), str(new_version), 1)
```
ShimonOhayon, line 92: @ilyam8 Avoiding further replacing of string after first hit (badges are in top of the README.md file)

ilyam8:
```suggestion
            data = data.replace(str(old_version), str(new_version), 2)
```

There are 2 occurrences of app/chart version in this "Artifact HUB" (`[name](url)`)

ilyam8: Thanks for highlighting this line btw 😄

ShimonOhayon: Good Catch !

.github/scripts/update_versions.py
```diff
@@ -0,0 +1,108 @@
...
+        FileHandler.write_chart(self.chart_data)
+
+    def get_app_version(self):
+        print(self.chart_data['appVersion'])
+
+    def set_app_version(self):
+        new_app_version = self.version
+        FileHandler.write_readme(self.chart_data['appVersion'],
+                                 new_app_version)
+        self.chart_data['appVersion'] = new_app_version
```
ilyam8, line 71:
```suggestion
        self.chart_data['appVersion'] = 'v' + new_app_version
```

Just noticed there is `v` prefix

https://github.com/netdata/helmchart/blob/6dc7e8fcdfc8c58815f3de7332de9335cd721ac0/charts/netdata/Chart.yaml#L28

ilyam8: Approving after the fix.

ShimonOhayon: Yes but both the current and new version will be supply with the `v` prefix . if you notice for `appVersion` i do string replace and not actual version bump, This value will come from [netdata repository agent bump  PR workflow](https://github.com/netdata/netdata/blob/master/.github/workflows/docker.yml#L130)

ilyam8: I mean not readme, but Chart.yml

ilyam8:
```diff
0 helmchart (issue-1506/Automate_version_bumping %)$ .github/scripts/update_versions.py set_app_version --version 1.99.0
0 helmchart (issue-1506/Automate_version_bumping *%)$ git diff
diff --git a/charts/netdata/Chart.yaml b/charts/netdata/Chart.yaml
index fa9960d..2b3579d 100644
--- a/charts/netdata/Chart.yaml
+++ b/charts/netdata/Chart.yaml
@@ -25,4 +25,4 @@ maintainers:
   - name: Oleksii Kravchenko
     email: kamikaze.ua@gmail.com
 icon: https://netdata.github.io/helmchart/logo.png
-appVersion: v1.31.0
+appVersion: 1.99.0
diff --git a/charts/netdata/README.md b/charts/netdata/README.md
index 7380a29..6ef98a5 100644
--- a/charts/netdata/README.md
+++ b/charts/netdata/README.md
@@ -1,6 +1,6 @@
 # Netdata Helm chart for Kubernetes deployments

-[![Artifact HUB](https://img.shields.io/endpoint?url=https://artifacthub.io/badge/repository/netdata)](https://artifacthub.io/packages/search?repo=netdata) ![Version: 3.7.1](https://img.shields.io/badge/Version-3.7.1-informational) ![AppVersion: v1.31.0](https://img.shields.io/badge/AppVersion-v1.31.0-informational)
+[![Artifact HUB](https://img.shields.io/endpoint?url=https://artifacthub.io/badge/repository/netdata)](https://artifacthub.io/packages/search?repo=netdata) ![Version: 3.7.1](https://img.shields.io/badge/Version-3.7.1-informational) ![AppVersion: 1.99.0](https://img.shields.io/badge/AppVersion-1.99.0-informational)

 _Based on the work of varyumin (https://github.com/varyumin/netdata)_.
```

ilyam8: Ok, it depends on `${{ github.event.inputs.agent_version }}`. If it has `v` feel free to close this converstion.

ilyam8 (approved):

.github/scripts/update_versions.py
ShimonOhayon: @ilyam8 No worries thanks for the review

Ferroin (dismissed):

charts/netdata/README.md
```diff
@@ -84,7 +84,7 @@ The following table lists the configurable parameters of the netdata chart and t
...
 | `replicaCount`                           | Number of `replicas` for the parent netdata `Deployment`                                                                                               | `1`                                                                                     |
 | `image.repository`                       | Container image repo                                                                                                                                   | `netdata/netdata`                                                                       |
-| `image.tag`                              | Container image tag                                                                                                                                    | Latest stable netdata release (e.g. `v1.31.0`)                                          |
+| `image.tag`                              | Container image tag                                                                                                                                    | Latest stable netdata release                                              |
```
knatsakis, line 87: Let's keep the table properly aligned (referring to the final `|` character)

knatsakis (changes requested):

ilyam8 (approved):

M4itee (approved):

knatsakis (approved):

PR #253, opened 2022-02-01
Add parent/child.envFrom

rickard-von-essen:
This allows adding environment variables from `Secrets` and `ConfigMaps`
created outside of this Chart.

E.g. to read _Netdata Cloud claim token_ from an existing secret `netdata` set this to:

```yaml
parent:
  envFrom:
    - secretRef:
        name: netdata
```

And create it with:
```shell
$ kubectl create secret generic netdata --from-literal="NETDATA_CLAIM_TOKEN=<token>"
```

Also ensure that `parent.claim.token` is empty.


charts/netdata/values.yaml
```diff
@@ -114,6 +114,13 @@ parent:
     ## To disable anonymous statistics:
     # DO_NOT_TRACK: 1
 
+  envFrom: []
+    ## E.g. to read Netdata Cloud claim token from an existing secret "netdata" set this to:
+    # - secretRef:
+    #     name: netdata
+    # And create it with: kubectl create secret generic netdata --from-literal="NETDATA_CLAIM_TOKEN=<token>"
+    # Also ensure that claim.token is empty
```
ilyam8, line 122:
```suggestion
    ## And create it with: kubectl create secret generic netdata --from-literal="NETDATA_CLAIM_TOKEN=<token>"
    ## Also ensure that claim.token is empty
```

We use double `#` for comments

charts/netdata/values.yaml
```diff
@@ -309,6 +316,13 @@ child:
...
```
ilyam8, line 324:
```suggestion
    ## And create it with: kubectl create secret generic netdata --from-literal="NETDATA_CLAIM_TOKEN=<token>"
    ## Also ensure that claim.token is empty
```

ilyam8 (approved):

ilyam8:
Thanks for the contribution, @rickard-von-essen 👍

M4itee (approved):
Thanks for that :)

PR #254, opened 2022-02-01
ci: do not run release GHA job on forks

ilyam8:
https://github.com/ilyam8/helmchart/runs/5018544354?check_suite_focus=true


M4itee (approved):
this should be the default option...

BillKalaitzis (approved):

PR #275, opened 2022-04-19
Child pod Probes host key

emiran-orange:
By default, netdata listens on 127.0.0.1.
With hostNetwork set to true in child, pod healthchecks can work if host key is set accordingly (e.g. `host: 127.0.0.1`)


ilyam8:
Hi, @emiran-orange. Thanks for the contribution. Can you elaborate on why we need that? When having 127.0.0.1 is better than the pod IP? Also, if 127.0.0.1 is better and it is the only possible value, we can set it automatically if `child.hostNetwork` is true.

emiran-orange:
@ilyam8 
When hostNetwork is true, the pod IP is the host IP and the kubelet performs the healthchecks against that IP.
As netdata listens by default on localhost only, the healtchecks fail inevitably.
An insecure alternative would be to make netdata listen on all interfaces but that is not an option for me especially because of using host network.
BTW the same issue occurs for parent pod but, as it does not us host network, having netdata listen on all interfaces is fine by me.

emiran-orange:
> Also, if 127.0.0.1 is better and it is the only possible value, we can set it automatically if `child.hostNetwork` is true.

To have the child healthchecks for in any case, regarding my previous comment, it would require:
- set host in healthchecks to 127.0.0.1 when hostNetwork is true
- add "bind socket to IP = 0.0.0.0" in global section when hostNetwork is false

Maybe a better way would be to set it as a default in values.yml and let people disable it whatever the reason (i.e. not condition it on `child.hostNetwork`) for flexibility.
What do you think ?

ilyam8:
> When hostNetwork is true, the pod IP is the host IP and the kubelet performs the healthchecks against that IP.

Correct.

> As netdata listens by default on localhost only, the healtchecks fail inevitably.

The default is to listen on all interfaces. Do you mean our netdata docker image has this default or where do we set it?

---

> set host in healthchecks to 127.0.0.1 when hostNetwork is true

Yes, I think this is what we need and I see no reason why we need this to be a configurable option.

The scenario when the probes fail is:

 - hostNetwork is true (default)
 - Netdata bind to localhost (non-default)
 - probes fails because they use podIP (==hostIP)

The following covers the case:

```yaml
          livenessProbe:
            httpGet:
              path: /api/v1/info
              port: http
              {{- if .Values.child.hostNetwork }}
              host: 127.0.0.1
              {{- end }}
```

The scenario:

 - hostNetwork is false (non-default)
 - Netdata bind to localhost (non-default)
 - probes fails httpGet is a wrong choice they don't do

---

If we want to cover all the cases we can switch from httpGet to exec and use [health.sh](https://github.com/netdata/netdata/blob/master/packaging/docker/health.sh)

emiran-orange:
Actually I think I misinterpreted the healthcheck failure on child pods and was misguided by some docs stating netdata was only bound to 127.0.0.1.
I had things fixed and it all works fine after all. Sorry for the fuss.

PR #276, opened 2022-04-19
Add nodes/metrics ressource in clusterRole

emiran-orange:
Required for go.d/k8s_kubelet when kubelet's insecure endpoint is disabled

### DIFF
--- a/charts/netdata/templates/clusterrole.yaml
+++ b/charts/netdata/templates/clusterrole.yaml
@@ -11,12 +11,12 @@ metadata:
 rules:
   - apiGroups: [""]
     resources:
-      - "pods"        # used by sd, netdata (cgroup-name.sh, get-kubernetes-labels.sh)
-      - "services"    # used by sd
-      - "configmaps"  # used by sd
-      - "secrets"     # used by sd
-      - "nodes"       # used by go.d.plugin (k8s_state)
-      - "nodes/metrics"       # used by go.d.plugin (k8s_kubelet)
+      - "pods"           # used by sd, netdata (cgroup-name.sh, get-kubernetes-labels.sh)
+      - "services"       # used by sd
+      - "configmaps"     # used by sd
+      - "secrets"        # used by sd
+      - "nodes"          # used by go.d.plugin (k8s_state)
+      - "nodes/metrics"  # used by go.d.plugin (k8s_kubelet)
     verbs:
       - "get"
       - "list"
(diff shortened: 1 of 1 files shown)


charts/netdata/templates/clusterrole.yaml
```diff
@@ -16,6 +16,7 @@ rules:
       - "configmaps"  # used by sd
       - "secrets"     # used by sd
       - "nodes"       # used by go.d.plugin (k8s_state)
+      - "nodes/metrics"       # used by go.d.plugin (k8s_kubelet)
```
ilyam8, line 19: Let's indent comments

emiran-orange: Good point sorry...
Fixed in e941655

charts/netdata/templates/clusterrole.yaml
```diff
@@ -11,11 +11,12 @@ metadata:
...
-      - "services"    # used by sd
-      - "configmaps"  # used by sd
-      - "secrets"     # used by sd
-      - "nodes"       # used by go.d.plugin (k8s_state)
+      - "pods"           # used by sd, netdata (cgroup-name.sh, get-kubernetes-labels.sh)
+      - "services"       # used by sd
+      - "configmaps"     # used by sd
+      - "secrets"        # used by sd
+      - "nodes"          # used by go.d.plugin (k8s_state)
+      - "nodes/metrics"  # used by go.d.plugin (k8s_kubelet)
```
ilyam8, line 19: Can this be covered by the following?

```yaml
  - nonResourceURLs: ["/metrics"]
    verbs:
      - "get"
```

ilyam8: Not a specialist, but if yes, I think it is better - more explicit that we need a non-resource /metrics endpoint.

emiran-orange: It can't as nodes/metrics are actual resources.
I tested it anyway (my other wrong PR lowered my self-confidence ;-)) but it does not work

ilyam8: Ah, you are right. Didn't know that, I guess I've found the relevant documentation - https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-authentication-authorization/#kubelet-authorization.

Let's then mention that it is when querying the HTTPS endpoint in the comment.

emiran-orange: Fixed in [fce91c7](https://github.com/netdata/helmchart/pull/276/commits/fce91c7de4d194eb28cfca6c9c4b6ef36fbf63cc)

ilyam8 (approved):

ShimonOhayon (approved):

ilyam8:
@emiran-orange, thanks!

PR #285, opened 2022-06-17
disable ML on children instances

ilyam8:
- we [enabled by default](https://github.com/netdata/netdata/pull/13158)
- we want to enable ML only on parent instances


ilyam8:
@andrewm4894 is this PR correct? Should ML be disabled on children instances by default?

PR #286, opened 2022-06-17
add K8s state Deployment

ilyam8:
This PR adds a new Netdata instance (Deployment) that [collects K8s cluster state](https://github.com/netdata/go.d.plugin/tree/master/modules/k8s_state#kubernetes-cluster-state-monitoring-with-netdata).

Added manifests:

- netdata-k8s-state (Deployment).
- netdata-conf-k8s-state (ConfigMap).
- netdata-k8s-state-varlib (PVC).

Why a new Deployment:

- 'k8s_state' collects the state of K8s objects not bound to nodes (e.g. Pods).
- We don't want to have a copy of it on every node (children - DaemonSet).
- Using the parent instance is not a proper solution too.

> **Note**: 
> - **New instance is disabled by default** because the latest stable Netdata has no k8s_state collector. I will enable it after the next release.
> - But we need to have this merged so we can use k8s_state in our infra (we use netdata/netdata:latest).

---

- effectively this PR adds no changes because k8sState instance is disabled - produced manifests before and after this PR are equal.
- I did test this PR with k8sState enabled by deploying it on my GKE cluster - works well.


Ferroin (dismissed):
Approving based on code inspection (the code appears to do what the PR description says it does) as I don’t have the required infra to properly test myself currently.

charts/netdata/templates/configmap.yaml
```diff
@@ -37,6 +37,26 @@ data:
...
+  name: netdata-conf-k8s-state
+  labels:
+    app: {{ template "netdata.name" . }}
+    chart: {{ template "netdata.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+data:
+  {{- range $name, $config := .Values.k8sState.configs }}
+  {{- if $config.enabled }}
+  {{ $name }}: {{ tpl $config.data $ | toYaml | indent 4 }}
```
M4itee, line 53:
```suggestion
  {{ $name }}: {{ tpl $config.data $ | toYaml | indent 4 | trim }}
```

this will give you:
```
data:
  go.d: |
      default_run: no
      modules:
        k8s_state: yes
```

instead of:
```
data:
  go.d:     |
      default_run: no
      modules:
        k8s_state: yes
```

M4itee (dismissed):
It works, I just have one suggestion.

charts/netdata/templates/deployment.yaml
```diff
@@ -159,3 +160,156 @@ spec:
...
+        app: {{ template "netdata.name" . }}
+        release: {{ .Release.Name }}
+        role: k8sState
+{{- with .Values.k8sState.podLabels }}
+{{ toYaml . | trim | indent 8 }}
+{{- end }}
+    spec:
+      securityContext:
+        fsGroup: 201
+      serviceAccountName: {{ .Values.serviceAccount.name }}
```
witalisoft, line 203: service account should have create set to `true`

charts/netdata/values.yaml
```diff
@@ -330,3 +330,130 @@ child:
...
+    # requests:
+    #  cpu: 4
+    #  memory: 4096Mi
+
+  livenessProbe:
+    failureThreshold: 3
+    periodSeconds: 30
+    successThreshold: 1
+    timeoutSeconds: 1
+  readinessProbe:
```
witalisoft, line 351: `readinessProbe` having the same values as `livenessProbe` make it useless

charts/netdata/values.yaml
```diff
@@ -330,3 +330,130 @@ child:
...
+  podLabels: {}
+
+  podAnnotationAppArmor:
+    enabled: true
+
+  podAnnotations: {}
+
+  dnsPolicy: ClusterFirstWithHostNet
+
+  varLib:
```
witalisoft, line 378: nit: I would omit `varLib`, because it is too enigmatic to end-user
```
  persistence:
    enabled: true
    ## Set '-' as the storageclass to get a volume from the default storage class.
    storageclass: "-"
    size: 1Gi
```

charts/netdata/templates/deployment.yaml
ilyam8: Sorry, can't get it. Can you reword, please?

charts/netdata/values.yaml
ilyam8: I agree. In addition, the defaults we have are not good IMO. However, we have them for all Netdatas:
 - parent
 - children

So I just copied them. Changing the values looks like a separate PR for me.

charts/netdata/templates/deployment.yaml
ilyam8: Ah, I see. Then I guess we could make it a requirement for the whole thing - 0 reasons to create k8sState instance if either rbac.create or serviceAccount.create is not true.

ilyam8: > service account should have create set to true

In theory, there could be a service account created manually/not from our helmchart - I guess that is the reason we have no if/else around it for [parent](https://github.com/netdata/helmchart/blob/53499b3e29cea279cb1e9585db1fd43e38a9bb19/charts/netdata/templates/deployment.yaml#L36)/[children](https://github.com/netdata/helmchart/blob/53499b3e29cea279cb1e9585db1fd43e38a9bb19/charts/netdata/templates/daemonset.yaml#L38-L39). If that is the case - we don't depend on `serviceAccount.create` here - a user can just set the name to "" if they want to use the default account.

witalisoft: ok, so if `serviceAccount.create` is set to true I would leave `serviceAccount.name` empty  and add named templates:
```
{{- define "serviceAccountName.netdata" -}}
{{- if .Values.serviceAccount.create -}}
    {{ default (include "netadata.fullname" .) .Values.serviceAccount.name }}
{{- else -}}
    {{ default "default" .Values.serviceAccount.name }}
{{- end -}}
{{- end -}}
```
any options inside `serviceAccount` won't make helm fail, because of the `default`

this is rather a suggestion

witalisoft (approved):

Ferroin (approved):

PR #293, opened 2022-07-26
feat(child): add `extraHostVolumeMounts` option

ilyam8:
That option allows mounting additional host volume mounts - this feature is needed (for instance) for the diskspace plugin. See https://github.com/netdata/netdata/discussions/13425.

Tested locally with `helm template -s templates/daemonset.yaml .` - w/ and w/o `extraHostVolumeMounts`.

@BillKalaitzis

- I think we need only HostVolumeMounts, so the option should say it explicitly.
- We need this option only for the child instance because it is the only instance that collects the host metrics.
- node exporter helm chart has the same option (extraHostVolumeMounts).


BillKalaitzis:
Hello @ilyam8, I was actually preparing to create a PR for adding the generic `extraVolumes` and `extraVolumeMounts`, as it was discussed in here: https://github.com/netdata/netdata/discussions/13425

I don't think we should limit our helmchart to only be able to mount `HostVolumeMounts` (and only on the `daemonset`).
Most helmcharts come with the generic `extraVolumes / extraVolumeMounts` directive, and I think we should do the same.
Even if we might not currently have any use cases for non hostPath volumes, I think our users will definitely find a way to benefit from them.

ilyam8:
@BillKalaitzis give me an example or two when we need extra not host mountpoints? I think we need to add this option when we need that, not now.

M4itee:
@ilyam8 I would like to be able to monitor `NFS`, `iscsi` or `fc` volume or any other read-write-many volume for that matter. Such volume I would mount to parent because I would not like to have multiplied metrics for it.

ilyam8:
@M4itee, I am not a specialist, but sounds like you need `extraVolumes / extraVolumeMounts` for the parent instance, not for the Daemonset. I still think that we need only extraHostVolumeMounts for the Daemonset.

BillKalaitzis:
Hello @ilyam8 ,

> * give me an example or two when we need extra **not host** mountpoints?
> * gime me an example or two when we need to mount extra volume mounts to parent/k8sState?


I think that each collector that requires any kind of path on its configuration is a candidate for requiring regular volume and volumeMounts.
For example, the `go.d web collector` (https://learn.netdata.cloud/docs/agent/collectors/go.d.plugin/pkg/web) requires a path to the CA/CERT/KEY if the user needs to monitor a URL that's signed by a self-signed certificate.
In that case, our users would probably need to mount a configmap/secret with the respective TLS certs on their child and/or parents.

Other use-cases could include the fact that someone might not want to use our helmchart generated configmap to store any sensitive information. In this case, a sensitive configuration (e.g postgres.conf), would need to be a secret and would be mounted as a regular Volume.

ilyam8:
Closing in favor of #294

PR #294, opened 2022-08-02
feat: add support for extraVolumes and extraVolumeMounts

BillKalaitzis:
Adds support for the generic `extraVolumes` and `extraVolumeMounts` directives on our helmchart.
The capabilities provided by them have come in handy on the following discussion items:
- https://github.com/netdata/netdata/discussions/13425
- https://github.com/netdata/netdata/discussions/13463


charts/netdata/values.yaml
```diff
@@ -333,6 +337,19 @@ child:
     rooms: ""
     url: "https://app.netdata.cloud"
 
+  extraVolumeMounts: []
+    ## Additional volume mounts for netdata child
+    ## E.g to mount all disks under / to be monitored via the diskspace plugin
+    # - name: root
+    #   mountPath: /host
+    #   readOnly: true
```
ilyam8, line 345: Default [mountPropagation](https://kubernetes.io/docs/concepts/storage/volumes/#mount-propagation) is None. I think it makes sense to set it to HostToContainer in this example (_E.g to mount all disks under / to be monitored via the diskspace plugin_) or at least mention the option. What do you think?

BillKalaitzis: Hey @ilyam8, that's a good catch! I've appended the suggested mountProgagation option in here: https://github.com/netdata/helmchart/pull/294/commits/cce91f8891dcdadf125046f6558f2856784cc65d

ilyam8:
@BillKalaitzis readme should be updated too 😄

BillKalaitzis:
> @BillKalaitzis readme should be updated too 😄

Done 😄

ilyam8 (approved):

M4itee (approved):

PR #305, opened 2022-08-15
update maintainers

ilyam8:
An attempt to keep the list up to date

### DIFF
--- a/charts/netdata/Chart.yaml
+++ b/charts/netdata/Chart.yaml
@@ -12,13 +12,7 @@ sources:
   - https://github.com/netdata/helmchart
   - https://github.com/netdata/netdata
 maintainers:
-  - name: Ilya Mashchenko
-    email: ilya@netdata.cloud
-  - name: Mateusz Bularz
-    email: mat@netdata.cloud
-  - name: Bill Kalaitzis
-    email: bill@netdata.cloud
-  - name: Witold Duranek
-    email: witold@netdata.cloud
+  - name: Netdata Cloud SRE team
+    email: cloud-sre@netdata.cloud
 icon: https://netdata.github.io/helmchart/logo.png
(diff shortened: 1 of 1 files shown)


charts/netdata/Chart.yaml
```diff
@@ -14,15 +14,11 @@ sources:
...
-  - name: Karuppiah Natarajan
-    email: karuppiah7890@gmail.com
-  - name: Oleksii Kravchenko
-    email: kamikaze.ua@gmail.com
+  - name: Mateusz Bularz
+    email: mat@netdata.cloud
+  - name: Bill Kalaitzis
+    email: bill@netdata.cloud
+  - name: Witold Duranek
+    email: witold@netdata.cloud
```
ilyam8, line 22: Probably a better way to put a team here, not individual people.

Ferroin: Agreed. Perhaps we should create an internal alias specifically for those maintaining the Helm chart?

ilyam8: I was thinking of just using the Cloud SRE team. I don't think that is so important we need a dedicated alias just for helmchart maintainers.

Ferroin: Well yes, but it’s not _just_ the Cloud SRE team.

ilyam8: Ok, maintainers is [optional](https://helm.sh/docs/topics/charts/#the-chartyaml-file). I think we can just remove the field. I think it is ok, maintained by Netdata.

ilyam8: @witalisoft added Cloud SRE team, please approve

BillKalaitzis (approved):

underhood (approved):

PR #341, opened 2023-03-06
fixing the name for netdata parent

M4itee:
Fixing netdata-parent name in cloud

### DIFF
--- a/charts/netdata/values.yaml
+++ b/charts/netdata/values.yaml
@@ -148,6 +148,7 @@ parent:
       path: /etc/netdata/netdata.conf
       data: |
         [global]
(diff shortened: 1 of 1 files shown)


charts/netdata/values.yaml
```diff
@@ -148,6 +148,7 @@ parent:
       path: /etc/netdata/netdata.conf
       data: |
         [global]
+          hostname = netdata-parent
```
ilyam8, line 151: Please update k8sState name too (it is Deployment, we should update its name too).

ilyam8 (approved):

dimko (approved):

PR #371, opened 2023-07-04
Remove /var/run/docker.sock mount

GabrielDyck:
Hi!

# Problem

In Kubernetes 1.24, dockershim and /var/run/docker.sock are deprecated in favor of CRI.

We have looked that netdata uses /var/run/docker.sock  to collect containers metadata, logs, and metrics  and is not possible to read it from helm values.

Please, give me your thoughts about it.
Thank you in advance.

Links related:
https://kubernetes.io/docs/tasks/administer-cluster/migrating-from-dockershim/
https://kubernetes.io/docs/tasks/administer-cluster/migrating-from-dockershim/migrating-telemetry-and-security-agents/
https://aws.amazon.com/es/blogs/containers/all-you-need-to-know-about-moving-to-containerd-on-amazon-eks/

### DIFF
--- a/charts/netdata/templates/daemonset.yaml
+++ b/charts/netdata/templates/daemonset.yaml
@@ -154,7 +154,7 @@ spec:
               readOnly: true
               mountPath: /host/proc
             - name: run
-              mountPath: /var/run/docker.sock
+              mountPath: {{ .Values.child.volumes.run.path }}
             - name: sys
               mountPath: /host/sys
             - name: os-release
@@ -224,7 +224,7 @@ spec:
             path: /proc
         - name: run
           hostPath:
-            path: /var/run/docker.sock
+            path: {{ .Values.child.volumes.run.path }}
         - name: sys
           hostPath:
             path: /sys
--- a/charts/netdata/templates/daemonset.yaml
+++ b/charts/netdata/templates/daemonset.yaml
@@ -153,8 +153,6 @@ spec:
             - name: proc
               readOnly: true
               mountPath: /host/proc
-            - name: run
-              mountPath: {{ .Values.child.volumes.run.path }}
             - name: sys
               mountPath: /host/sys
             - name: os-release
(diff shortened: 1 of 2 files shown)


witalisoft:
based on the Kubernetes version we should pass the correct path for container runtime plus the option to overwrite

ilyam8:
> We have looked that netdata uses /var/run/docker.sock to collect containers metadata

This is only for Docker containers when installed directly on the host. Netdata uses the K8s API to collect container metadata when installed inside a K8s cluster ([cgroup.plugin](https://github.com/netdata/netdata/tree/master/collectors/cgroups.plugin#monitor-cgroups-cgroupsplugin)).

- `/var/run/docker.sock` is not really used.
- there is no reason to mount `/run/containerd/containerd.sock` because Netdata doesn't use it.

So if Netdata installation doesn't fail with current values I think we can keep it as is.

GabrielDyck:
> 
Thanks for your prompt reply.

Given that this helm chart install Netdata in K8S, this leads me to think if it is necessary to mount this volume.

In https://kubernetes.io/docs/tasks/administer-cluster/migrating-from-dockershim/migrating-telemetry-and-security-agents/#identify-docker-dependency there is a command to get agents which uses docker engine. 

Perhaps, keeping the "run" volume  with docker.sock could lead to confusion.

ilyam8:
Yes, I think we can remove that mount.

GabrielDyck:
Great.

I've removed the run volumen from the daemonset and I've installed the chart in a minikube cluster. Containers are running and panels in the UI show values in every panel. 

Let me know if I can help you checking anything else.

ilyam8 (approved):

GabrielDyck:
Hi!

We need one more approval. @witalisoft Do you agree with this? 

Let me know if I can help you with another thing.

dimko (approved):

PR #396, opened 2023-11-17
add namespace property to all namespaced objects

rbjorklin:
Using the helm built-in `.Release.Namespace` as per: https://helm.sh/docs/chart_template_guide/builtin_objects/

### DIFF
--- a/charts/netdata/templates/configmap.yaml
+++ b/charts/netdata/templates/configmap.yaml
@@ -5,6 +5,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: netdata-conf-parent
+  namespace: {{ .Release.Namespace }}
   labels:
     app: {{ template "netdata.name" . }}
     chart: {{ template "netdata.chart" . }}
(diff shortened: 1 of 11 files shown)


ilyam8:
Hey, @rbjorklin. Thanks for the contribution! But can you please explain why this is needed? Did you have a problem installing the current chart? I see that [this topic is somewhat controversial and not really a best practice [according to one of helm devs](https://github.com/helm/helm/issues/5465#issuecomment-473942223). Also, I don't see it is recommended in [Helm Best Practices](https://helm.sh/docs/chart_best_practices/conventions/) docs (at least I couldn't find).

rbjorklin:
Hey @ilyam8! Thanks for taking the time to dig up those links.
This might boil down to a limitation in our internal tooling but without these changes the chart does not apply correctly. We are currently using an internal fork of the chart with these changes and I was hoping to upstream them.

ilyam8:
I think it is ok to change, the only thing I have in mind (not a k8s specialist) that would break is 

- generating with helm (and don't provide `--namespace`, it will explicitly set the namespace to default)
- pipe the result to `kubectl` with `--namspace` != default

e.g.

```bash
helm template . | kubectl apply --namespace someNamespace -f -
```

but I am not sure why someone would want to do it.

witalisoft (approved):

M4itee (approved):
It basically changes nothing in grand scheme of things in terms of functionality. If it helps some people to avoid some kind of problem/bug - let's go.

Issue #402, opened 2024-01-16
avoid child open port and fix liveness probe on public worker nodes


alexfouche:
On a k8s cluster with public worker node, the child daemonset will be available publicly on port 19999, since it runs with `hostNetwork: true`

From the Securing Netdata pages, and since there is a Netdata parent accessible from a ClusterIP service, it is possible to avoid that by setting :

```yaml 
child:
  configs:
    netdata:
      data: |
        [web]
          mode = none
(...)
```

Unfortunately this breaks the liveness and readiness probes.

So i am using 

```yaml 
child:
  configs:
    netdata:
      data: |
        [web]
          bind to = localhost:19999
(...)
```

The issue i have is that the Helm values.yaml does not allow to set the `httpGet.host:` field, so that it can be set to `127.0.0.1`

I tested by editing the daemonset livenessProbe and readinessProbe manually, and it works:

```yaml
livenessProbe:
      httpGet:
        host: 127.0.0.1
        path: /api/v1/info
        port: http
        scheme: HTTP
(...)
```

Shouldn't it even be set by default for the Netdata child to not listen on all interfaces, since there will be the Netdata parent it streams to anyway ?

ilyam8:
> Unfortunately this breaks the liveness and readiness probes.

You can use livenessProbe exec - `netdatacli ping`. We use it for [Docker health checks](https://github.com/netdata/netdata/blob/master/packaging/docker/health.sh) with disabled web server.

BirknerAlex:
@ilyam8 That's one solution but unfortunately there is no option to override the probes for the child processes in the helm chart, see #433 which at least adds the host option to it.

ilyam8:
@BirknerAlex I think we need to implement it because it will allow to use `web. mode = none`. #433 fixes only `web bind to = localhost:19999`

BirknerAlex:
Yap, thats the reason why I wrote "partically" in the PR. I don't wanted to touch the helm chart that much (due lack of time) but at least provide one possible solution for that kind of issue (public IP nodes).

PR #403, opened 2024-01-16
Issue https://github.com/netdata/helmchart/issues/402

alexfouche:
Bind Netdata child to localhost interfaces by default and fix livenessProbe and readinessProbe

### DIFF
--- a/charts/netdata/templates/daemonset.yaml
+++ b/charts/netdata/templates/daemonset.yaml
@@ -134,6 +134,7 @@ spec:
               protocol: TCP
           livenessProbe:
             httpGet:
+              host: 127.0.0.1
               path: /api/v1/info
               port: http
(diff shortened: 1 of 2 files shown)


witalisoft:
@alexfouche Thanks for the contribution. 

Consider providing these changes as an option, because when someones have their own `netdata.conf` it will override the `bind to` and the health check will fail.

M4itee:
@alexfouche thanks for your work!
It does need the change. We cannot merge it and expect that everybody will have the same setup. The default configuration with such change should always point to the old one whenever possible.

M4itee:
I'm closing this PR due to incompatibility with a current default setup. @alexfouche please reopen if the changes we proposed are fine with you.

alexfouche:
I opened a new pull request with accommodation from your previous comments at https://github.com/netdata/helmchart/pull/406

PR #405, opened 2024-02-08
reorganise the chart with one folder per component

jouve:
a benefit is that `config/checksum` annotation is now calculated exactly for its own configmap, so changing a parent config won't restart the daemonset for example.


M4itee:
@jouve Thanks for this PR. It looks much better and solves a real problem. We need to test this on our own to make sure that it will not break the existing setups.

witalisoft:
LGTM good job 👍

witalisoft (approved):

M4itee (approved):

PR #406, opened 2024-02-09
 Issue https://github.com/netdata/helmchart/issues/402 #403 

alexfouche:
Provide a mean to fix liveness and readiness probes when forcing the Netdata child web UI to bind to localhost.

Accordingly to remarks of [previous pull request](https://github.com/netdata/helmchart/pull/403/), i did not changed the default which opens Netdata webUI to everybody, even though i find it a very questionnable default.

Also updated README for the new properties, and explanation on how to configure Helm values when Kubernetes worker nodes are on a public network or accessible publicly

### DIFF
--- a/charts/netdata/README.md
+++ b/charts/netdata/README.md
@@ -134,11 +134,17 @@ The following table lists the configurable parameters of the netdata chart and t
 | `parent.enabled`                              | Install parent Deployment to receive metrics from children nodes                                                                                       | `true`                                                                                  |
 | `parent.port`                                 | Parent's listen port                                                                                                                                   | `19999`                                                                                 |
 | `parent.resources`                            | Resources for the parent deployment                                                                                                                    | `{}`                                                                                    |
+| `parent.livenessProbe.method`                 | How the liveness probe is initiated. Either "httpGet" or "exec"                                                                                        | `httpGet`                                                                               |
+| `parent.livenessProbe.httpGet`                | Liveness probe "httpGet" definition                                                                                                                    | `{path: /api/v1/info, port: http}`                                                      |
+| `parent.livenessProbe.exec`                   | Liveness probe "exec" definition                                                                                                                       | `{command: [netdatacli, ping]}`                                                         |
 | `parent.livenessProbe.initialDelaySeconds`    | Number of seconds after the container has started before liveness probes are initiated                                                                 | `0`                                                                                     |
 | `parent.livenessProbe.failureThreshold`       | When a liveness probe fails, Kubernetes will try failureThreshold times before giving up. Giving up the liveness probe means restarting the container  | `3`                                                                                     |
 | `parent.livenessProbe.periodSeconds`          | How often (in seconds) to perform the liveness probe                                                                                                   | `30`                                                                                    |
 | `parent.livenessProbe.successThreshold`       | Minimum consecutive successes for the liveness probe to be considered successful after having failed                                                   | `1`                                                                                     |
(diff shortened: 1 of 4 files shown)


ilyam8:
I am not sure we need those changes:

- Concern that Netdata children/k8state instances listening on 0.0.0.0 is valid. It is not needed and I agree that default should be localhost.
- I don't understand the problem from [this comment](https://github.com/netdata/helmchart/pull/403/#issuecomment-1899851110) - yes, it will be overwritten. But if one decides to overwrite `netdata.conf` he is responsible for its content.

I think we need just change `bind to` for child and k8s state instances.

ilyam8:
@alexfouche thanks for your work and sorry I didn't comment earlier in #402. I think #402 is ok.

ilyam8:
If for some reason we want to keep `bind to` as it is, still we don't need changes from this PR. 

We'd need:

- add `host: 127.0.0.1` to the probes (part of #402)
- @alexfouche will change `bind to` using custom `netdata.conf` in his installation.

alexfouche:
> I don't understand the problem from https://github.com/netdata/helmchart/pull/403/#issuecomment-1899851110 - yes, it will be overwritten. But if one decides to overwrite netdata.conf he is responsible for its content.

indeed i did not understand the comment either, and that is why i also customized the Netdata child probes to probe 127.0.0.1 to make sure nothing was broken.
I simply supposed what was meant by the comment was Netdata child not accessible from all IP when nodes are on a private network and possibly with no parent, could be a breaking change for some users... who knows
Maybe @M4itee had some case in mind

alexfouche:
I added probe customisation for parent and k8state only to uniformize the Helm values configuration for all probes. It may be unnecessary or overkill, but well, it allows overriding the probes, some people may like it

alexfouche:
I do not know which of the https://github.com/netdata/helmchart/pull/403 or this one https://github.com/netdata/helmchart/pull/406 is better or best for user

403 is simpler and straight to the point. Child does not listen by default on all interfaces and probes probe accordingly. (indeed that should be also the case for the k8state)

406 does not include any change to the current behavior, but you have to figure out yourself it is accessible or read the last part of the readme to notice it, and customize both child, k8state netdata.conf and probes configurations to close access.

It may not be a problem for most Kubernetes cluster, but there are still many managed Kubernetes providers where nodes are not on private networks. Note that the problem is the same for all `hostNetwork` pods exposing metrics such as Prometheus Node Exporter, Traefik, ...

witalisoft:
@ilyam8 
> I don't understand the problem from https://github.com/netdata/helmchart/pull/403/#issuecomment-1899851110 - yes, it will be overwritten. But if one decides to overwrite netdata.conf he is responsible for its content.

the k8s probes were hardcoded to `127.0.0.1` so if you want to bind netdata to a specific IP (not 127.0.0.1) you won't get chance. So my point was to provide in helm value an option with the `host` for the k8s probes.

witalisoft:
@alexfouche 

I would personally keep the changes from https://github.com/netdata/helmchart/pull/403, but without specifying [this](https://github.com/alexfouche/charts-netdata/blob/dacf28baa4450eab6343d59a462a9c019a649adc/charts/netdata/values.yaml#L304) and having helm value options for setting the `host` parameter for the `livenessProbe` and `readinessProbe`. If you don't specify these helm values it will not change the current behavior (no breaking changes from the scenario presented https://github.com/netdata/helmchart/pull/406#issuecomment-1938264648), but if you set those values you can achieve your goals.

M4itee:
closing due to lack of response

Issue #445, opened 2024-10-10
Add ability to set hostname for netdata-parent and netdata-k8s-state kubernetes pods


jamesharr:
We have several Kubernetes clusters deployed and I'd like to be able to tune just the hostname via a simple helm value.

For example:
```sh
# First cluster
kubectl config use-context cluster1
helm --namespace netdata-system upgrade netdata netdata-netdata \
  --reset-then-reuse-values \
  --set parent.hostname='cluster1-parent' \
  --set k8sState.hostname='cluster1-k8s-state'

# Second cluster
kubectl config use-context cluster2
helm --namespace netdata-system upgrade netdata netdata-netdata \
  --reset-then-reuse-values \
  --set parent.hostname='cluster2-parent' \
  --set k8sState.hostname='cluster2-k8s-state'
```

### Current Work-around

Our current work-around is to override the Helm values `k8sState.configs.netdata.data` and `parent.configs.netdata.data` fields. This works, but it's both awkward since it's putting a multi-line config file into a single helm variable. It's also less than ideal for long term maintenance - If netdata updates the helm chart, for example, adds new plugins that need to have some basic configuration, then a user would need to track down and tune the config instead of taking advantage of out-of-the-box defaults.

PR #447, opened 2024-10-16
feat: add parent/k8sState hostname value

ilyam8:
Fixes: #445


witalisoft (approved):

jamesharr:
> I'm not sure we need to set `deployment.spec.template.spec.hostname` as well 🤷‍♂️ Not doing that in this PR.

I think you made the right call. I'd advise against setting `deployment.spec.template.spec.hostname` unless a compelling need emerges.
1. Since you're setting the hostname in the netdata config file, it achieves the desired behavior in #445 and ignores the container's hostname
2. In my k8s experience, it's exceedingly uncommon to set the container's hostname. Doing so might be confusing when troubleshooting something. When troubleshooting, I typically expect the hostname to be the container's name if I shell in.

Thanks for the quick patch, I appreciate it.

PR #459, opened 2024-12-31
rm service discovery sidecar for Netdata v2.2+

ilyam8:
[service-discovery sidecar](https://github.com/netdata/agent-service-discovery) is no longer needed as its functionality is built into go.d.plugin.

This change was announced in [v2.1.0 deprecation notice](https://github.com/netdata/netdata/releases/tag/v2.1.0#v210-deprecation-notice).

### DIFF
--- a/charts/netdata/templates/_helpers.tpl
+++ b/charts/netdata/templates/_helpers.tpl
@@ -41,8 +41,8 @@ Return the target Kubernetes version
 {{/*
 The function determines if the Netdata version is less than 3.0-0.
 */}}
-{{- define "netdata.verLess3" -}}
-{{- if semverCompare "<3.0-0" .Chart.AppVersion -}}
+{{- define "netdata.verLess22" -}}
+{{- if semverCompare "<2.2-0" .Chart.AppVersion -}}
 {{- "true" -}}
 {{- else -}}
 {{- "" -}}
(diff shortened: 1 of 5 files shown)


witalisoft (approved):

M4itee (changes requested):
I'm ok with the change but please keep `yaml` extensions the same - `k8s.yml` stands out from other `.yaml`'s

ilyam8:
The existing configuration has a `.yml` extension, so I created a new one with the same extension.

https://github.com/netdata/helmchart/blob/master/charts/netdata/sdconfig/child.yml

charts/netdata/templates/_helpers.tpl
```diff
@@ -38,6 +38,17 @@ Return the target Kubernetes version
 {{- default .Capabilities.KubeVersion.Version .Values.kubeVersion -}}
 {{- end -}}
 
+{{/*
+The function determines if the Netdata version is less than 3.0-0.
+*/}}
+{{- define "netdata.verLess3" -}}
+{{- if semverCompare "<3.0-0" .Chart.AppVersion -}}
```
ilyam8, line 45: This will change if we decide not to do a major release. See [this Slack discussion](https://netdata-cloud.slack.com/archives/CNLAM2B19/p1735663645012739) for details.

ilyam8:
@M4itee can you answer to https://github.com/netdata/helmchart/pull/459#issuecomment-2568839549?

ilyam8:
Changed version check to < 2.2 as we decided not to do a major release.

M4itee (approved):

PR #471, opened 2025-02-07
adding conditional to claim rooms

M4itee:
Since we have dynamic rooms it makes sense for this parameter to be optional.


ilyam8 (approved):
Whether 'NETDATA_CLAIM_ROOMS' is empty or not defined, the outcome will be the same. Wil it not?

witalisoft (approved):

PR #475, opened 2025-03-05
update clusterrole: add cronjobs and jobs

ilyam8:


M4itee (dismissed):

charts/netdata/templates/clusterrole.yaml
```diff
@@ -29,6 +29,13 @@ rules:
       - "get"
       - "list"
       - "watch"
+  - apiGroups: ["batch"]
+    resources:
+      - "cronjobs"
```
M4itee, line 34: I would probably also add jobs while we are at it.

ilyam8: added

M4itee (approved):

PR #482, opened 2025-04-09
Add Startup Volume Permission Enforcement

### LINKED ISSUE #449: Feature Request | Enforce Volume Permissions
Dear Netdata,

Could I please request a feature whereby we can enforce permissions for a volume.

Here is a rough example
```
volumePermissions:
## @param netdata.volumePermissions.enabled Enable init container that changes the owner and group of the persistent volume(s) mountpoint to `runAsUser:fsGroup`
        ##
        enabled: true
        ## @param netdata.volumePermissions.image.tag Init container volume-permissions image tag (immutable tags are recommended)
```

Thank you.

theeternalrat:
Resolves #449 . This ads the ability run the parent pod as a select uid/gid, and run an init container that will set file permissions appropriately on the mounted directories.

### DIFF
--- a/charts/netdata/templates/parent/deployment.yaml
+++ b/charts/netdata/templates/parent/deployment.yaml
@@ -60,7 +60,7 @@ spec:
       {{- end }}
       {{- if .Values.parent.volumePermissions.enabled }}
       {{- with .Values.parent.volumePermissions }}
-        - name: volumePermissions
+        - name: volume-permissions
           image: "{{ .image.repository }}:{{ .image.tag }}"
           imagePullPolicy: {{ .image.pullPolicy }}
           securityContext:
--- a/charts/netdata/templates/parent/deployment.yaml
+++ b/charts/netdata/templates/parent/deployment.yaml
@@ -58,45 +58,7 @@ spec:
           resources:
 {{ toYaml .Values.sysctlInitContainer.resources | indent 12 }}
       {{- end }}
-      {{- if .Values.parent.volumePermissions.enabled }}
-      {{- with .Values.parent.volumePermissions }}
-        - name: volume-permissions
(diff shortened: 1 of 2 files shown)


witalisoft:
Thanks for the PR

Could you tell us what kind of use case it solves ?

btw there are some files inside the containers that are bound to that builtin user netdata (uid: 201)
how will this work after overriding the `runAsUser` or/and `runAsGroup` ?

theeternalrat:
@witalisoft I can't speak for the original author of #449, but my use-case is that I am running Netdata on an AWS EFS backed volume. If the U/ID of the container (201) does not match the ID of the filesystem, the Netdata process quits when it tries to `chown` the data directories. EFS squashes the file permissions for all operations, but it doesn't allow any `chown` operation on the files. Should any process attempt to `chown`, it is not allowed. Using this feature, I was able to set the container ID to match the EFS volume, allowing the Netdata startup checks to pass.

I cannot definitively say what effect might be for those specified files, but I have been running these changes in my cluster for the last day with seemingly normal operation.

witalisoft:
Sounds reasonable, but having another use case introduces another piece of code to maintain, especially if it's a divergence from what we expect to have. 

So what I would prefer is to put the placeholder for the `initContainers` in `templates/parent/deployment.yaml` like this:

```yaml
...
{{ toYaml .Values.parent.extraInitContainers | indent 8 }}
      containers:
...
```

Now you get the flexibility, and we're not stuck with another use case for testing the helm chart.

The rest of what you presented like `securityContext` is ok.

theeternalrat:
@witalisoft I've made the requested change. It'll now be up to the user to defined their own volume-permissions container if they so choose.

witalisoft (approved):

M4itee (approved):
looks good. Thanks @theeternalrat

PR #532, opened 2026-06-18
Add missing otel signal viewer plugin for the otel Agent

ralphm:
As the default configuration explicitly disables "new" plugins, all plugins that are not explicitly enabled will be disabled. Since the set of enabled plugins for use within k8s has been created a while ago, new plugins such as the `otel-signal-viewer` plugin are not mentioned, even though we do expect them to be there.

### DIFF
--- a/charts/netdata/values.yaml
+++ b/charts/netdata/values.yaml
@@ -1040,6 +1040,7 @@ netdataOpentelemetry:
           cgroups = no
           enable running new plugins = no
           otel = yes
+          otel-signal-viewer = yes
           journal-viewer = yes
           slabinfo = no
           perf = no


ralphm:
@M4itee We may want to add some other plugins as well, like `network-viewer`. This problem will continue to pop up when we add new plugins.

We currently have `enable running new plugin = no`, which means 'disable all by default', so  all the explicit `= no` entries we have now are basically no-ops. Possibly, we could have that setting as 'yes' and then only disables the plugins we do not want?

mtblz:
we can, the goal was for this endpoint to ONLY host logs (no overhead)

ralphm:
Per out-of-band discussion, this change should be sufficient for the time being. Please review.

witalisoft (approved):

mtblz (approved):

mtblz:
Mind you Vasilis said that this is not going to be needed in the future (now it still is)

PR #534, opened 2026-06-25
docs: explain the OpenTelemetry logs setup in the chart README

mtblz:
## What

Adds a concise **Collecting logs with OpenTelemetry** section to the chart README, explaining the two components involved in OpenTelemetry log ingestion and why both must be enabled.

## Why

`netdataOpentelemetry` and `otel-collector` both exist in `values.yaml`, but the README never explained their relationship — that both are required for logs to appear, and that the bundled collector is an optional, replaceable convenience. This previously required manual explanation to users.

## Changes

- **`templates/netdata-README.md.gotmpl`** — new `### Collecting logs with OpenTelemetry` subsection (placed after *Persistent volumes*), covering:
  - **`netdataOpentelemetry`** — the OTLP receiver (port `4317`) that stores and displays logs; it does not collect them.
  - **`otel-collector`** — the optional `DaemonSet` (one pod per node) that reads node-local container logs and forwards them over OTLP; replaceable by any OTLP-capable pipeline.
  - An end-to-end log-flow diagram.
- **`charts/netdata/README.md`** — regenerated via `./generate-documentation.sh`.

## Notes

- Documentation only — no chart logic, templates, or values changed.
- The generated README stays in sync with the template, so the *"documentation is up-to-date"* CI check passes.

### DIFF
--- a/charts/netdata/README.md
+++ b/charts/netdata/README.md
@@ -2343,6 +2343,92 @@ otel-collector DaemonSet  (one pod per node, needs host log access)
 netdata-otel:4317  →  stored and shown in the Netdata UI
 ```
 
+#### Securing the endpoint with TLS
+
+By default the `netdata-otel` receiver listens on port `4317` in **plaintext** — TLS is **disabled**
+(`tls_cert_path`, `tls_key_path`, and `tls_ca_cert_path` are unset in
+`netdataOpentelemetry.configs.otel.data`). The steps below turn it on with a self-signed
+certificate. TLS affects **both sides**: the receiver must serve the certificate, and every client
+(including the bundled `otel-collector`) must be switched to TLS, or it will stop delivering data.
+
+**1. Generate a self-signed certificate and key** (Linux, `openssl`):
+
+```
+openssl req -x509 -newkey rsa:4096 -nodes \
+  -keyout tls.key -out tls.crt -days 365 \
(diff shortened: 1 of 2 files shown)


charts/netdata/README.md
```diff
@@ -2313,6 +2313,37 @@ In case of `child` instance it is a bit simpler. By default, hostPath: `/var/lib
 in: `/var/lib/netdata`. You can disable it, but this option is pretty much required in a real life scenario, as without
 it each pod deletion will result in a new replication node for a parent.
 
+### Collecting logs with OpenTelemetry
+
+Netdata can ingest, store, and visualize your cluster's container logs through OpenTelemetry.
+This relies on two components — **both disabled by default** and **both required** for logs to
```
witalisoft: `both required` ? later on there is an info that `otel-collector` is an optional subchart

witalisoft (commented):
It is worth mentioning, especially if you want to push the logs yourself, that by default there is no TLS enabled for the OTEL endpoint in `netdataOpentelemetry`

charts/netdata/README.md
mtblz: it is explained more later but I will reword it a bit.

witalisoft (approved):

ilyam8 (approved):

